Issue #1, opened 2019-10-01
Implement get_current_tree in levels/levels.py


bthayer2365:
This function should parse the current git branches and return it in a json format. Look at the code for details. The level_json function should be useful.

PR #3, opened 2019-10-05
Implemented get_current_tree

wertydoo:
Builds a tree of the current repository consisting of all commits, branches, tags, and the HEAD. In addition, commits lists all parents, branches lists all target commits, and tags lists all target commits.

### DIFF
--- a/gitgud/operations.py
+++ b/gitgud/operations.py
@@ -119,15 +119,56 @@ def create_tree(commits, head):
 
 def get_current_tree():
     # Return a json object with the same structure as in level_json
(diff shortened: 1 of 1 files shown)


bthayer2365:
Thanks for the code, Brain. It does exactly what I need it to do!

 You traverse some of the commits a few extra times, but that's a good opportunity for someone interested in algorithms to optimize the code. It's only when there's a lot of commits that this would be an issue anyways. I'll open an issue for anyone who's interested!

bthayer2365:
fixes #1

PR #6, opened 2019-10-05
Added level 'intro merging'

Fiss-and-Fuse:


bthayer2365:
Hey @Fiss-and-Fuse , thanks for the PR! All the code and spec files look good and are ready to merge. Before that, I have a few suggestions for the instructions.

Since this is still the intro, I think it's good if we handhold a bit more and make it clear exactly what the user is supposed to type in. The solution commands would be

"git branch bugFix"
"git checkout bugFix"
"git gud commit"
"git checkout master"
"git gud commit"
"git merge bugFix"

Can you weave those into the instructions, including the quotes so the user can see what the commands are?

For example, the 3rd line might look like this:
```
Make a new branch called bugFix using "git branch bugFix"
```
Follow a similar pattern with the rest of the instructions and this will be good to go. Thanks again for the PR!

bthayer2365:
Thanks @Fiss-and-Fuse. You're merged!

PR #7, opened 2019-10-05
Added level 'intro branching'

parkmichael777:


bthayer2365:
Hey Michael, thanks for the PR! Everything looks good, just a few updates and I'll be happy to merge.

Can you delete these two lines

```
Hit enter to advance.
```
and
```
Great! You just advanced through the first part of the first set of instructions!
```

The user will already have done that in the first level, so they won't need to do that again.

Also, we also already say "let's start using git!" in the commits challenge too, so can you change that last line to something like "Now that we know how to make a branch, we can start seeing how git operates in the real world!"

parkmichael777:
Hey Ben,

I modified those three things and pushed.

Thanks,
Michael Park

bthayer2365:
Looks good! You're merged!

parkmichael777:
Alright, thank you!

PR #12, opened 2019-10-07
Optimized get_current_tree so every commit is visited only once.

wertydoo:
Fixed bad python so that every node in the commit tree is only visited once. Please note that this was only tested for a small set of cases and more robust unit testing should be considered in the future.


bthayer2365:
Looks good Brian, thanks for updating this. Code should be as efficient as it gets now.

PR #19, opened 2019-10-13
Create help command

### LINKED ISSUE #8: Add `git gud help <command>`
This shouldn't be that hard. ArgParse parsers have a `.show_help()` method. Just add a subparser that uses the `help` command then display the appropriate help message.

Pseudocode should be something like `git gud help <command>` and then run `<command>_parser.show_help()`

Relevant file is gitgud/__main__.py

Ishan1742:
fixes #8

### DIFF
--- a/gitgud/__main__.py
+++ b/gitgud/__main__.py
@@ -26,20 +26,26 @@ class GitGud:
 
         # TODO Display help message for subcommand when it fails.
         # ie `git gud load level1 challenge1 random-input` should have output similar to `git gud load --help`
+        
+        '''
+        Use "git gud help" to print helps of all subcommands.
+        "git gud help <command>" prints the description of the <command> but not help.
+        Longer descriptions can be added later
+        '''
 
-        help_parser = subparsers.add_parser('help', help='Show help for commands') 
-        start_parser = subparsers.add_parser('start', help='Git started!')
-        status_parser = subparsers.add_parser('status', help='Print out the current level')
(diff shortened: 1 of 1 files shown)


bthayer2365:
This looks good. Thank you!

I have a four small things for you to change, shouldn't be hard.

First, This is more a stylistic preference of mine than anything, but I think it's good for consistency. Can you change the multiline comment to use hashes and be on the lines immediately before the parsers. And change the add longer descriptions to be a TODO, so:

```
# Use "git gud help" to print helps of all subcommands.
# "git gud help <command>" prints the description of the <command> but not help.
# TODO Add longer descriptions
help_parser = ...
```

Next, Instead of self.subparser_copy, let's use `self.subparsers` and change the `subparsers = ...` to `self.subparsers = ...`. You'll need to change all the references to `subparsers` as well.

Keep the "No such command exists." That's good. if you could, on the same line, also output the command that doesn't exist to make it clear. So
`git gud help asdf`
prints
`No such command exists: "asdf"`

And if we get a KeyError, instead of printing a list of commands, why not just print what we would as if we ran "git gud help". It would have pretty formatting and provide more information.

After that, we're good. To make a list out of it:
- [ ] Change comment to use # and be on line before parsers
- [ ] Change usages to `self.subparsers`
- [ ] Output invalid command
- [ ] Instead of listing commands, run normal help prompt

bthayer2365:
Everything looks good! Merging!

Issue #11, opened 2019-10-07
Implement `git gud show-tree` to display tree in `.spec` format


bthayer2365:
It would be great for the user to have a simple way to display the current tree. Simply display the current tree in the spec format this project uses when specifying levels.

The spec format can be seen in the level directories ie 'levels/merging/_octopus/test.spec'

To do this, you may also want to implement `operations.get_topology()`, which may be used for other functions, but this is not necessary.

bthayer2365:
@parkmichael777 Can you work on this with the command alias you showed me from before that showed the git tree? You'll need to make use of GitPython, but essentially you're just running the same command that you showed me. We don't need to do it in the `.spec` format.

parkmichael777:
Sure, I can take a look at it!

PR #20, opened 2019-10-13
Created "git gud show-tree"; shows commit history

parkmichael777:


gitgud/operations.py
```diff
@@ -43,6 +43,10 @@ def add_and_commit(self, name):
 
         return commit
 
+    def show_tree(self):
+        tree = self.repo.git.log('--graph', '--oneline', '--all', '--decorate=short')
```
bthayer2365: From https://git-scm.com/docs/git-log:

The default option for --decorate is short, so we can remove it.

Also, I think show_tree implies that it'll be the one showing it, so I think it should handle that. The quick solution to that would be to move the print statement there, but......

Git itself shows pretty colors when running that command. Can you see if you can avoid using a print statement and have git do the printing so it keeps all the pretty colors?

parkmichael777: I'll get rid of =short, but I'm going to have to leave "--decorate"; otherwise, the references won't show up (unlike manually typing "git log --graph --oneline --all" where '--decorate' is implied for some reason) 

Yeah, I originally thought if I left out the print command, it would execute as if I had wrote "git log etc." But, if I don't turn it into a string and print it (just run "self.file_operator.repo.git.log(<flags>)) nothing will happen. There will be a short delay from some processing, but nothing will print to console.

bthayer2365:
Hmm. Could be different versions of git leading to the --decorate issue.
I'll have to see for myself.

As for printing to the console, you could either keep working with
gitpython or you can see if you can use the command line and call git
directly. This shouldn't be a problem some git log doesn't change the
state. If you can't come up with anything, it's fine to leave it as is.

parkmichael777:
I've managed to get it working with colors but you need to import another module, "subprocess."

bthayer2365:
That's fine. Subprocess is built in, so there's no extra dependencies. I'll
test it out later.

parkmichael777:
Okay, I pushed it

bthayer2365:
Closes #11

PR #21, opened 2019-10-13
Nicer output for `git gud levels` 

### LINKED ISSUE #9: Make `git gud levels` look pretty
I'd like to have some pretty output for `git gud levels` that displays the name of all levels in git gud. 

Right now it's just a series of print statements with the number of challenges. If it could be something that looked nicer and displayed more information, that would be great.

The method is already there in \_\_main\_\_.py, so all that you'd need to do is find a way to make it pretty.

sahansk2:
fixes #9 
There is prettier padding for the levels being printed, with them being lined up in columns.
The word "challenge"  is made plural or singular depending on how many challenges are in the level.
The challenges for each level are listed under each level, spaced so that the colons line up.
(Might want to squash and merge, the various commits in the branch just change comments -- sorry)

### DIFF
--- a/gitgud/__main__.py
+++ b/gitgud/__main__.py
@@ -161,10 +161,12 @@ class GitGud:
         for level in all_levels.values():
             # TODO Make pretty
             # TODO Add description
+            # 10 characters for the short IDs. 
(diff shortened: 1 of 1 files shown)


sahansk2:
Sounds good -- I saw a similar code snippet elsewhere using count, so I've refactored that as well.

bthayer2365:
Alright, I see it. All looks good. Thanks!

PR #22, opened 2019-10-14
KeyError when using `git gud load`

### LINKED ISSUE #18: KeyError when using `git gud load`
I'd like to have this fail gracefully instead of just showing the stack trace for the KeyError.

The KeyErrors happen when providing a level name or a challenge name that doesn't exist.

Instead of showing the KeyError, we should provide a message that says "Level X doesn't exist" or "Challenge Y doesn't exist" and something like "To view challenges/levels, use `git gud challenges` or `git gud levels`"

rblcoder:
fixes #18

### DIFF
--- a/gitgud/__main__.py
+++ b/gitgud/__main__.py
@@ -190,16 +190,24 @@ class GitGud:
 
     def handle_load(self, args):
         self.assert_initialized()
+        if args.level_name in all_levels:
+            level = all_levels[args.level_name]
 
-        level = all_levels[args.level_name]
-
-        if args.challenge_name is not None:
-            challenge = level.challenges[args.challenge_name]
+            if args.challenge_name is not None:
+                if args.challenge_name in all_levels[args.level_name].challenges:
+                    challenge = level.challenges[args.challenge_name]
+                    challenge.setup(self.file_operator)
+                    self.file_operator.write_challenge(challenge)
+                else:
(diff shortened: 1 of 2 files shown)


rblcoder:
@bthayer2365 Please see the changes.

bthayer2365:
I tested it. Works perfectly! But... you actually wrote more code than you need to.

`get_challenge` will never actually encounter a KeyError because it is reading from a file that Git Gud itself is writing to. Since the "write_challenge" method only accepts challenge objects as input, it will always write a valid level name. If it doesn't, then something's actually wrong and we will want the stack trace. 

On the other hand, user input will have typos every once in a while, so we shouldn't act like something's wrong if we expect it to happen.

Also, since we're displaying the user's input, I think it makes sense to put that in quotes, so instead of having
```"Level {} ..."```
we should do
```"Level \"{}\" ..."```

So, that leaves two things for you to do:
1. Change the code in `get_challenge` back to what it was before
2. Add quotes around user input

PS Since the two sets of changes are in different files, you can just checkout an old revision of operations.py to get the old `get_challenge` code back
`git checkout HEAD~1 -- gitgud\operations.py`

rblcoder:
@bthayer2365 Please see the changes.

bthayer2365:
Good stuff! Thanks for the PR!

PR #24, opened 2019-10-17
Added first three challenges in ramp_up level + stub of fourth level

parkmichael777:
TODO: 
1. Add the last ramp_up level
2. For _rel_ref2, need to devise a way to show node 6 using git gud show-tree even though it is unreferenced.
3. For _rel_ref2, need to figure out why git gud can't setup with HEAD pointing at a commit node.


parkmichael777:
Okay, so the reason that the head doesn't get assigned to the commit node during setup is because it is set equal to "2" in parse_spec. Then, at line 117 of operations.py, the program searches in vain for a branch that has the name "2" to make the HEAD. We need another for loop below this where we cycle through a dictionary like {commit names:commit hashes}; if the head matches a name in the dictionary, it will checkout the corresponding hash (e.g. (operator).repo.git.checkout(ab12)). Also, as long as the default branches (i.e. master, bugFix) are not set as a number, I don't think there will be a conflict because create_tree is only called during setup right? Anyways, let me know if this breaks the architecture, and whether there is some easier way of identifying the corresponding commit hashes. I remember you talked about using tags in order to save unreferenced commit nodes—maybe we can extend that system to record the head node?

bthayer2365:
Hey @parkmichael777, the first thing you said was correct. The current version of the code only looks for branches since that's all I thought of originally. All we need to do is support the case of the branch not being found and then instead looking for a commit with that name. We can assume the branch names are never going to be numbers since we are the ones controlling that.

The only difference with what you said is that we can keep track of GitPython's commit objects instead of just the raw hashes.

gitgud/levels/ramp_up/__init__.py
```diff
@@ -0,0 +1,11 @@
+from collections import OrderedDict
+
+import pkg_resources
+
+from gitgud.levels.util import BasicChallenge
+
+all_challenges = OrderedDict()
+all_challenges['detaching'] = BasicChallenge('detaching', pkg_resources.resource_filename(__name__, '_detaching/'))
+all_challenges['rel refs 1'] = BasicChallenge('rel_refs_1', pkg_resources.resource_filename(__name__, '_rel_refs1/'))
```
bthayer2365: The command line won't accept names with spaces. I think we should change this name to relrefs1 and relrefs2. Similarly, it is my intent that these folders should be able to be used as python modules, which should have names with only lowercase letters and no underscores. Numbers are fine, and leading underscores are fine. Leading underscores signal that it is only meant to be used by the module and not by people importing the module.

gitgud/levels/__init__.py
```diff
@@ -1,13 +1,15 @@
...
 from gitgud.levels.intro import all_challenges as intro_challenges
+from gitgud.levels.ramp_up import all_challenges as ramp_up_challenges
 from gitgud.levels.merging import all_challenges as merging_challenges
 
 from gitgud.levels.util import Level
 
 all_levels = OrderedDict()
 
 all_levels['intro'] = Level('intro', intro_challenges)
+all_levels['ramp_up'] = Level('ramp_up', ramp_up_challenges)
```
bthayer2365: Change ramp_up to rampup for same reason as stated in my other comment

bthayer2365:
Comments on your TODOs:

### 1. 
We're going to need a way to check if the commit is being reverted, that is something that is currently ill-defined in the spec files. This is the kind of situation where we can create a python file and extend the BasicChallenge class to give us what we need. We can write a test.spec file to test if  C0, C1 and C2 are all created correctly, and then the commit denoted C2', which reflects a reverted commit, we can check manually for it. We can also change `test_level` to allow for extra branches to be present. Currently, if there are extra branches, then the test will fail, but we can change it so that if we have an extra "pushed" branch that isn't checked for (only master/local are checked for correctness) then we should be okay. In our test function, we can then separately check to make sure pushed has file 2 deleted and that its parent is commit 2.

### 2.
#34 will solve this

### 3.
I addressed this in my original response to your comment

parkmichael777:
Ok, I'll get these done Saturday or if I have time this week!

parkmichael777:
The current pushed version (94b3) is missing the test/setup for reversing (pending discussion about revert syntax). Everything else is done.

bthayer2365:
Hey @parkmichael777, thanks for making all the changes I requested. Now we can move onto the more in-depth discussion of the level instructions, and with a lot of additions come a lot of comments. My comments are sorted by challenge. My comment on challenge 4 (reversing changes) applies to everything. In short, I think we need to separate this PR into 4 different ones so we can have a more focused discussion. Use `git rebase -i` to make the new branches. 

#### 1. 
I don't think detaching head is quite clear what to do. Can you add more specific instructions? We don't quite have the benefit of displaying the goal state yet, so we'll need to have more specific instructions.

#### 2.
We didn't learn that we only need 4 digits of the commit hash.

I'd also call it "another way" instead of a "more efficient way". Relative refs are used when it's just a few commits back, but for example, when I'm looking though the git logs to see the commit message, I'd rather not take the time to count back how many commits that is when I can read that the hash is `8636...`. In that case, reading the hash is more efficient.

`^` and `~` are not quite how you explained them. `^` is shorthand for `^1`, and `~` can be used as shorthand for `~1`. `^` is used to say which parent to go to. `^1` goes to parent 1, which in the case of a normal commit, is just the parent. `^2` will go to the merged commit. If there are more parents, like in an octopus merge, then `^3` and so on can be used, but there's no need to mention that in this level. Maybe in a future level, but for now, I think it's safe to keep it out. Similarly, `~` and `^` can be stringed together. `~~~` is possible, just like `^^^` is possible. Similarly, different sequences can be strung together like `~^~` or `~^2~5`, which would get the merge commit that happened prior to the current commit, access the other commit that was merged into the original branch and go five back from there.

Also, unfortunately, on windows (but not on git bash), `^` is the character that allows you to let a command take up mulitple lines. Instead, on windows we use `^^`, which escapes the character and converts it to an actual `^`. This should be noted.

The instructions explain stuff about relative refs but don't explain what to do to solve the level, so can you add that?

#### 3.
`setup.spec` and `level.spec` are a little wacky and don't really correspond to the instructions, please change those. Either clarify in the instructions what you mean or change the specs to align with the instructions more.

This also works with branches -> this also works with branches and tags, and will work any other reference

Again, instead of saying X commits "above", use the language I described above

I think you can remove this whole line: `A useful application of the tilde operator is the command "git branch -f <branch_name> HEAD~<num>". This command allows you to move a certain branch to a certain commit node <num> above the current node.` That isn't the most common use of this, but it certainly can be done and we should teach them how to do that, but still one thing at a time. You should open up a new issue to make a new level based on git branch -f <branch_name> HEAD~<num>. Another challenge that might go in the same level as that is using the git reset --hard HEAD~<num>, which would modify the current branch. Add those both to the issue. For future reference, "above" is an ambiguous term, you should say `<num> commits into the history/in the past` or `the previous <num>th commit`

#### 4.
I think you should move the work that you've done on this challenge and open up a different PR since working on this challenge will take longer given the stuff we need to work on for it. It would probably make sense to do this with every level honestly and open up more pull requests, so if you could do that, that would be nice. You can make use of the interactive rebase command (`git rebase -i <newbase>`) If there are commits where you edit multiple challenges, you'll also need to enter edit mode. It should be good practice, and honestly would make for a good level! See if you can do it, and then open up pull requests for each individual level and then reference this PR and we can go from there. It would be much easier than trying to have all these conversations at once.

parkmichael777:
I'm probably going to push this off until Hacknight, but I'll try to change what I can before then.

bthayer2365:
Closing. This is now split up into multiple pull requests, one for each challenge. 

#66 #67 #68 #69

PR #36, opened 2019-10-19
Run commands even with random extra input.

### LINKED ISSUE #33: `git gud <command> oaiwthont  23q858ingg2` should run as if we entered `git gud <command>`
Right now, when there's random extra input, git gud won't run any command, but it would be good if it still worked.

For example `git status i1h5y98agnvaguhewai` still outputs the status of the git repo. Likewise, if someone runs a command like `git gud load intro committing 32o85yoijsghoisegh`, then we should behave as if they ran `git gud load intro committing`. 

This shouldn't be too hard. We just have to keep removing arguments from the input list until we get something that works, or we'll just get down to `git gud` and we'll output the help message anyways.

The current parsers will handle bad input as opposed to extra input. For example, `git gud load intro notalevel` will be handled by `handle_load` and state that "notalevel is not a level".

Relevant function is `parse` in `gitgud/__main__`

sahansk2:
Fixes #33 .
Digging through the [docs](https://docs.python.org/3/library/argparse.html#the-parse-args-method), I found argparse has a function "parse_known_args()" that will handle bad arguments quite nicely: it returns a tuple, the first element being the Namespace, and the second is a list object with all the extraneous arguments.

As a bonus, if we ever wanted to do something with those extra arguments in the future, we could easily do that.


bthayer2365:
Wow. That was even simpler than I thought. Thanks for researching how to do this.

Could you just change one thing? I think it's more pythonic to avoid using indices if we can. To fix this, I think we can change the first line from `args = self.parser.parse_known_args()` to `args, _ = self.parser.parse_known_args()`. Since `parse_known_args` returns a tuple, Python will automatically be able to expand that and be able to set it equal to two variables. In this case, we don't need the second value, so we put it in a dummy variable `_`, which is the conventional way to say that we aren't going to use the value.

With that, we can change the other occurrences of `args[0]` to `args`, making the code easier to read.

bthayer2365:
Also, just a comment, I see you've merged my code into yours through a rebase.

If you just wanted to test out making a pull request, that's perfectly fine. If you just wanted to use github to merge my code, that's fine too, but there are also a few ways to do it from the command line:

```
git remote add ben https://github.com/bthayer2365/git-gud.git
```
then
```
git pull ben master
```
or, if you instead had already made changes and wanted to avoid a merge, you could have done either of the following:
```
git fetch ben
git rebase ben/master
```
or more succinctly
```
git pull --rebase ben master
```
Git pull is a shorthand for fetch/merge or, in the second case, because we have the --rebase flag, fetch/rebase.

sahansk2:
Sorry I wasn't able to implement this fix faster -- I had been studying for midterms (and still am -- I should be more active after Wednesday).
But yep, I intended on making sure I had a relatively recent version of the code before implementing things, so thanks for telling me about that!

bthayer2365:
No worries. Take your time, we're all busy. 

Also, I think it might not have been clear before, I'm assuming you thought that by `_`, I meant fill in the blank, but I meant to literally have the name be a single underscore. 

See here: https://www.datacamp.com/community/tutorials/role-underscore-python#IV

sahansk2:
Oops, this should fix it -- I actually tried to iterate through the bad arguments and print them to the user before I committed, but the _ looked ugly so I changed the variable name to badargs, and then I then scrapped the entire thing without changing the variable name back.

bthayer2365:
No problem, it's appropriate to give no output to the user in this situation. If it becomes a problem, we can think about adding some output.

If you were to output the changes, then you were right to change it.`_` is meant to say that we aren't ever going to use the variable, but if we're outputting the bad arguments, then we are actually using the variable, so we do want a descriptive name.

Anyways, that was it. Thanks for the changes!

Issue #31, opened 2019-10-18
Remove `git gud progress` and replace with `git gud load next`


benthayer:
People read progress as if it is the noun, when I originally meant it to be the verb. That is confusing, so I think it would be good to make loading the next level a special case of the `load` subcommand.

We'll also need to make sure we change the instructions to match this change, so nobody types in `git gud progress` expecting to load the next level

Relevant methods are `handle_progress` and `handle_load` in `gitgud/__main__.py`

TSSlade:
I'd like to try to take this on.

PR #56, opened 2019-10-20
Implemented naïve fix for git gud progress --> git gud load next

TSSlade:
Fixes issue #31


TSSlade:
Ran into some issues locally with what appeared to be tabs failing to convert to spaces. Can't tell from here whether that persists in the code that got pushed.

bthayer2365:
See #58

Issue #23, opened 2019-10-14
Change the data structure for all_levels and all_challenges to allow indexing


bthayer2365:
all_levels is defined in levels/\_\_init\_\_.py
all_challenges is defined in levels/intro/\_\_init\_\_.py and in levels/merging/\_\_init\_\_.py

Currently, they're both implemented using `OrderedDict`, but that's not as useful as it could be. Dicts allow us to access levels/challenges by name and preserving the order allows us to know which one comes after the next (duh).

Currently, we run into situations, where we need to run
```
level1 = next(iter(all_levels))
challenge1 = next(iter(level1.challenges))
```

but this is hard to read, so I would like that same code to look like
```
level1 = all_levels[0]
challenges = level1.challenges[0]
```

OrderedDict doesn't let us do that because it uses a linked list internally, which allows for fast insertion, deletion and iteration, but makes accessing elements by index very challenging. I'd like to use a different data structure that allows for both indexing like a list and key/value lookups like a dictionary. 

There should be something in Python's standard library that we can use, so we should not have to implement our own. If for some reason there isn't anything in the standard library, you can comment here and we can figure something else out.

After changing the data structure, we should go and change all the code that uses `next(iter(...))`

Ishan1742:
I am not sure if there is a data structure in the standard library but we might be able to do it using hashing. We hash the level string to a index and at that index we store the values for challenges. 

When we have a level_name we can compute it's hash and look it up at that index.
Or we could just iterate through all the indexes when we need to.

The issue here is that a string can be hashed to the same index but since we are the only ones creating the levels we can set the hash values ourselves instead of an algorithm. 

dictionary for levels
\-
| "intro" : 0
| "merging" : 1
\-

list of list for challenges

\-
| 0 -> [challenge_1, challenge_2, challenge_3]
| 1 -> [challenge_1, challenge_2]
\-

bthayer2365:
Hey @Ishan1742 , thanks for the suggestion. If we're going to do that, I think it would be best to wrap that in our own data structure, so the underlying mechanism would be your idea, and then we add the `__getitem__` method to the `Level` class so we don't have to worry about accessing the data structures ourselves.

For example, we should be able to do `intro['committing']` and `intro[0]`, and they would return the same challenge. `type(intro)` would also be equal to `Level`.

In order to do the same thing to get access to each levels, we might have an `AllLevels` object that we can define in `gitgud/levels/util.py` and use in a similar fashion.

Together, we might have:
[...]

PR #66, opened 2019-10-23
Added Level "Reverting"

### LINKED ISSUE #91: Add reverting notation to spec files
The conversation has been taking place as part of #66, which is adding a level that shows the user how to revert. The idea is to list ~~`-1`~~ `1-` in the spec file if that commit reverts commit 1.

parkmichael777:
fixes #91


bthayer2365:
This PR comes from #24 

We can continue the discussion about this and what to do about finding undone changes here

gitgud/levels/rampup/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+
+import pkg_resources
+
+from gitgud.levels.util import BasicChallenge
+
+all_challenges = OrderedDict()
+all_challenges['detaching'] = BasicChallenge('detaching', pkg_resources.resource_filename(__name__, '_detaching/'))
+all_challenges['relrefs1'] = BasicChallenge('relrefs1', pkg_resources.resource_filename(__name__, '_relrefs1/'))
+all_challenges['relrefs2'] = BasicChallenge('relrefs2', pkg_resources.resource_filename(__name__, '_relrefs2/'))
+all_challenges['reversing'] = BasicChallenge('reversing', pkg_resources.resource_filename(__name__, '_reversing/'))
```
bthayer2365: Can you remove all challenges except `reversing`?

gitgud/levels/rampup/_reversing/instructions.txt
```diff
@@ -0,0 +1,19 @@
+Sometimes you may find that you made changes that you aren't happy with. Wouldn't it be great if there was a simple way to undo all your changes and return to the last commit? Here are two ways!
+>>>
+The commands "git reset" and "git revert" both allow you to rollback changes to an older commit. While "git reset" is used for the local branch, "git revert" is used for public branches.
```
bthayer2365: Can we add a brief explanation of what differentiates the two commands at a technical level?
git reset throws away the commit you're working on while git revert preserves the history while still undoing changes.

Also, instead of using the word public, can we change the description to this: `"git revert" is typically used for remote branches that other people are likely to have based their work off of.`

parkmichael777: Changed the "public" part to what you requested and added a brief explanation.

gitgud/levels/rampup/_reversing/instructions.txt
```diff
@@ -0,0 +1,19 @@
...
+>>>
+Let's start with "git reset". There are three modes: --soft, --mixed(default), and --hard.
+>>>
+The "git reset --soft <commit_ref>" is not used often. The branch pointer will move to the specified <commit_ref> (can use hash, HEAD^, HEAD~<num>, etc!). All the changes you've made since that commit will be left as staged ("Changes to be committed" in git status).
+>>>
+The "git reset --mixed <commit_ref>" is the default mode of git reset. In other words "git reset <commit_ref>" implies the mixed mode. The functionality is similar to soft mode—the branch pointer will be moved to <commit_ref> and changes will be preserved. The only difference is that your changes will be unstaged (will need to use "git add").
+>>>
+The third option is the most extreme (but may be the most useful). When you want to rollback the clock completely to a certain commit: use "git reset --hard <commit_ref>". This command will annihilate any changes you've made since the specified commit. Essentially, using this command will reset your repository to exactly what it was at the time of the specified commit. Be careful with this option!
```
bthayer2365: So I think we should start by reiterating that git reset throws away the commit, and then we can introduce mixed and soft together for the situation of preserving work and then telling them to use hard when they want to throw away all their changes

Examples would probably be useful, so here's the kind of thing I think we should highlight:

A typical use case for git reset --hard is when you've made changes to the wrong branch. After you move the changes, you can bring the original branch to it's original spot. You might also be working on a branch and then you make some changes that you don't like. If you don't want to throw away those changes, you can `git branch oldwork` `git reset HEAD~2`. Things like that
The situation for mixed and soft are similar, basically, you've made changes that you want to keep, but you're not sure which part of them you want to keep. Mixed is used when you mostly don't want to keep them, soft is used when you mostly do want to keep them.

There's probably a more eloquent way of saying something like that that you can weave into the instructions

gitgud/levels/rampup/_reversing/instructions.txt
```diff
@@ -0,0 +1,19 @@
...
+>>>
+I mentioned "git reset" isn't the only revision command.
```
bthayer2365: Given the length of the instructions, it might be best to remove this line to make things more concise

gitgud/levels/rampup/_reversing/instructions.txt
```diff
@@ -0,0 +1,19 @@
...
+>>>
+The second method is "git revert". Using "git revert" will create a new commit that is the same as the one you wish to revert to. Assume we've made changes on top of the HEAD node that we don't like anymore. Using "git revert HEAD" will create a new commit that is the same as the original version of HEAD.
```
bthayer2365: Git revert doesn't quite work like this. Play around with it to see what I mean. It'll remove the changes in the commits that you specify, so if you have files 1, 2, and 3 and you do git revert HEAD, and HEAD added file 3, then we'll have files 1 and 2 with 4 commits adding 1, 2, 3, then removing 3 again.

We can also do things like `git revert HEAD~2^..HEAD` to revert a sequence of commits.

gitgud/levels/rampup/_reversing/instructions.txt
```diff
@@ -0,0 +1,19 @@
...
+>>>
+Now, that seems redundant doesn't it? However, that's precisely why revert is important! By being redundant, we are keeping track of the changes we made in the edited version of HEAD. Therefore, if we decided we actually needed those changes, we can return to them. This is unlike reset, which modifies the commit history (you wouldn't be able to recover the reset data).
```
bthayer2365: I don't know if I'd call it redundant so much as verbose, can you change this?

I'd also mention the added benefit of other people not having to go back to a version of history where the changes aren't made yet. Instead, git knows that changes were reverted and it can replay those changes on top of any work that other people have done

bthayer2365:
Copied from #24:

We're going to need a way to check if the commit is being reverted, that is something that is currently ill-defined in the spec files. This is the kind of situation where we can create a python file and extend the BasicChallenge class to give us what we need. We can write a test.spec file to test if C0, C1 and C2 are all created correctly, and then the commit denoted C2', which reflects a reverted commit, we can check manually for it. We can also change test_level to allow for extra branches to be present. Currently, if there are extra branches, then the test will fail, but we can change it so that if we have an extra "pushed" branch that isn't checked for (only master/local are checked for correctness) then we should be okay. In our test function, we can then separately check to make sure pushed has file 2 deleted and that its parent is commit 2.

bthayer2365:
That earlier comment is just for reference. After some thought, I think it would make the most sense to represent reverted commits with a negative, so if we have files 1,2,3 and we want to revert the commit that adds file 2, we add another commit as `-2`. 

This seems logical and should be consistent. We'll need to update the parser. Let me know if you agree that this is the best thing to do

parkmichael777:
I've modified the instructions and added a new test and setup. In theory, it works if the "-#" syntax works. Also, I think that's good syntax, you should proceed with that idea!

gitgud/levels/rampup/_reversing/instructions.txt
```diff
@@ -1,19 +1,17 @@
...
+The second method is "git revert". Using "git revert" will create a new commit that will undo the changes that were made in a specified commit. For example, suppose I commit three files, 1, 2, and 3(<-HEAD). Using "git revert HEAD" will create a fourth commit where there are only files 1 and 2. In other words, the changes made at HEAD (commit 3) will be undone. Similarly, I could have used "git revert HEAD^" and this would have created a fourth commit with only files 1 and 3.
 >>>
-I mentioned "git reset" isn't the only revision command.
+Notice that in the previous example the commit with files 1, 2, and 3 still exists in the commit history! Git revert allows you to keep track of the commits prior to the revert. Therefore, if we decided we actually needed those changes, we can return to them. If you recall, this is unlike reset which abandons the working commit.
```
bthayer2365: Another important thing is that other people won't be able to see if you've deleted those commits, so git will just assume that they're the ones that made the changes and still make the unwanted changes. Reverting will undo the changes for everyone since they'll see the changes as a new commit, which can then be processed by git.

gitgud/levels/rampup/_reversing/instructions.txt
```diff
@@ -1,19 +1,17 @@
...
 >>>
-Now, that seems redundant doesn't it? However, that's precisely why revert is important! By being redundant, we are keeping track of the changes we made in the edited version of HEAD. Therefore, if we decided we actually needed those changes, we can return to them. This is unlike reset, which modifies the commit history (you wouldn't be able to recover the reset data).
->>>
-Wow! You learned a lot (hopefully), prove yourself by completing the challenge!
+Wow! You learned a lot (hopefully), prove yourself by completing the challenge! Reverse all changes on the "local" branch and reverse changes made in commits 3 and 5 on the "remote" branch. Assume that the "local" and "remote" branches are as their names imply.
```
bthayer2365: You should tell them to remove commits 3 and 5 "in that order" since only one of the orderings is going to be correct.

bthayer2365: Just looked at `git revert 1..5` Since that  reverts 5 then 3, we should specify 5 then 3 as the order. It would also make sense to specify that `git revert x..y` works backwards somewhere in the instructions

parkmichael777: Oh, I want them to use the sequential revert

gitgud/levels/rampup/_reversing/setup.spec
```diff
@@ -1 +1,7 @@
-
+1 (master)
+2 : 1 (local)
+3 : 1
+4 : 3
+5 : 3
+6 : 4 5 (remote)
+local
```
bthayer2365: What do you think about this change:
Let's change local to be master and have nothing refer to commit 1. We can then revert the changes, fast forward to the state of the remote and revert the changes from there. Therefore, the correct sequence of commands is going to be
```
git reset HEAD~
git merge remote
git revert 1..5
```

bthayer2365: I'd like to hear what you think about this. I'm not sure if it's the best way to do it. It might be more logical to do it the way we currently have it. (Although, I still think we should just make the local branch master)

gitgud/levels/rampup/_reversing/test.spec
```diff
@@ -1 +1,9 @@
+1 (master, local)
+3
+4 : 3
+5 : 3
+6 : 4 5
+-5
+-3 (remote)
```
bthayer2365: Do you think you can work on making this syntax available by changing the parser code?

gitgud/levels/rampup/_reversing/instructions.txt
bthayer2365: So, you're thinking 
```
git revert 5
git revert 3
```
?

parkmichael777: I want them to use git revert HEAD~3..HEAD^2.

gitgud/levels/rampup/_reversing/instructions.txt
parkmichael777: Added

gitgud/levels/rampup/_reversing/instructions.txt
bthayer2365: We don't have any way of preventing them from doing things different ways right now, but that is the same net effect

gitgud/levels/rampup/_reversing/setup.spec
parkmichael777: Hmmm, I like it better when it's two separate tasks because it allows you to more clearly contrast the two commands.

bthayer2365: Okay, that's fine then.

gitgud/levels/rampup/_reversing/instructions.txt
parkmichael777: I specify to do it using one command in the instructions, the only way they could do it with one command is using the above command which forces that order. That being said, it's true that we can't prevent them from doing it using two commands. Although at the same time, if they did it with two commands and got the order wrong, they would think it's because they didn't do it with one command->then they would do it with one command and get it right :P

gitgud/levels/rampup/_reversing/test.spec
parkmichael777: sure

parkmichael777: Hmm do you think we need to make a "git gud revert" that names the revert commit for you? The test parser picks up "-#" just fine, but in the level the commit name is whatever message you write for the revert. (i.e. "Revert 5 \n\nThis reverts commit <hash>")

bthayer2365: So I think the best way to do this is to look at the commit object itself and just recognize it as deleting file 5 for `-5` and 3 for `-3`, etc. It doesn't matter that it's a revert. A revert is nothing special other than that git is able to construct the commit itself

bthayer2365: I honestly think we should do that for normal commits too, but it's obviously much easier to just look at the commit name.

gitgud/levels/rampup/_reversing/instructions.txt
```diff
@@ -0,0 +1,17 @@
...
+>>>
+Finally, one last thing! There is a shortcut to reverting consecutive commits! For example, suppose we commit five files separately and we want to remove the changes made in commits 2 through 4 (i.e. remove files 2, 3, and 4). Use this syntax: "git revert <older_ref>..<newer_ref>". The older ref is non-inclusive; therefore, it should be the parent of the oldest commit to be reverted. The newer ref is inclusive; meaning it is the newest commmit to be reverted. All commits in-between will also be reverted. For our example: "git revert HEAD~4..HEAD^". This will revert the child of HEAD~4 (= C2) through HEAD^ (=C4).
```
bthayer2365: Can we make a note about ordering here: 

"Also, since each of the changes are likely to build off of each other when specifying a sequence of commits like that, it's only natural to undo them in reverse order, so running "git revert HEAD~4..HEAD^" will revert C4, C3, then C2"

gitgud/levels/rampup/_reversing/setup.spec
```diff
@@ -0,0 +1,7 @@
+1 (master)
+2 : 1 (local)
```
bthayer2365: Change local to master and remove master

gitgud/levels/rampup/_reversing/instructions.txt
parkmichael777: Added

gitgud/levels/rampup/_reversing/setup.spec
parkmichael777: Changed

bthayer2365: Can you make sure master is checked out too?

parkmichael777: done

bthayer2365:
Great, looks like the only thing left is the modifications to make the `-3` stuff work

parkmichael777:
I'm not sure how to do this, but I have some pieces. 

My plan:
In line 156 of util.py, we check if test commit names are present in level commit. Add an elif that detects a '-' using array slicing. In this block, check if there is a file with the same name as commit_name[1:] (a number of variable digits). If yes, do something; if not throw false. Here's what I don't understand: where do I get the names of files that are in the repository? Don't I need an Operator object?

Another problem: I said above that if we find that the file is deleted we need to do something. From here on I'm not sure what is going on. It looks like the section starting from 168 checks if the length of the parents and the parent ID's are the same. But this goes back to the problem of having no control over the revert commit names, we can't check the ID part if we don't know the names.

bthayer2365:
So you're right, we don't have control over the commit names for reverts. The same is true for reverts. If you look at the first section of the code in `test_level`,  I do a very hacky job of finding the names of each of the merge commits, and I'll need to change this. The test_level code is really only responsible for checking equality and shouldn't have the responsibility of assigning names to each of the commits, which is why you don't see any `Operator`s in there.

Take a look at `BasicChallenge`. It uses the file operator to get the current tree and to parse the spec files. You'll see that the methods `parse_spec`, `level_json` and `file_operator.get_current_tree` are used. I don't think you'll need to modify `parse_spec` or `level_json` since we have control over the names there, but you'll probably need to modify `file_operator.get_current_tree`. You'll have access to the commit objects there and can change their names to be whatever you want. To figure out the name, you could either parse it to see if the name contains "Revert" and then look for the number. You can just change `Revert "4"` to `-4`, etc. 

Eventually, I think we'll want to actually look at the diff and check for the files that are changed, but that would work best if we have a larger PR working on doing that for normal commits, reverts, and for merges. Merges will be a little more complicated, but we can ignore that and the rest of this paragraph for now.

bthayer2365:
@parkmichael777 how's it going? Do you need any help?

parkmichael777:
@bthayer2365 Hey! I haven't tried any of your latest suggestions yet—all my work is due on Wednesday, and I've got an exam tomorrow. I'll probably look at this on Thursday or next Tuesday.

bthayer2365:
Okay, no problem. Good luck on your exam!

bthayer2365:
I opened this up as a new issue - #91. If you start working on it, open up a new pull request and reference that issue.

bthayer2365:
Just made a note on #91 that we should be doing `1-` for reverting 1 instead of `-1` so that we can use the name as a tag as well.

bthayer2365:

parkmichael777:
Hey, so do you want me to change the spec syntax and __init__.py for this PR and then do the parsing changes in the other?

bthayer2365:
The purpose of different pull requests in different branches to separate work that should be separated. If you want to work on another level that has reverting then you should do it in a different pull request, but otherwise it's fine to do it here

For the changes to `__init__.py`, keep them here since it's part of this level

bthayer2365:
Also to resolve conflicts use `git merge master` from this branch. It should make it clear what needs to happen

parkmichael777:
No, I mean for tasks left on this PR do you only want me to do those two things (change reverting/test.spec and __init__.py) or is there more for this PR?

parkmichael777:
Ohh, nevermind lol. I added the new syntax and __init__.py. Working on the parsing changes now

gitgud/operations.py
```diff
@@ -162,7 +181,7 @@ def get_current_tree(self):
             'target': target,
             'id': 'HEAD'
         }
-
+        print(tree)
```
bthayer2365: Can you get rid of this?

gitgud/operations.py
```diff
@@ -108,6 +108,21 @@ def create_tree(self, commits, head):
         if head_is_commit:
             self.repo.git.checkout(commit_objects[head])
 
+    # Revert detection helper function
+    @staticmethod
+    def find_Revert(commit_name):
```
bthayer2365: change this method definition to `parse_name(commit_msg)` since it's not just looking for the revert stuff, it's also handles other cases - aka parsing. It's also taking in the message, not the "name". 

Also, as a note, the first line of a commit message is what's considered to be the "title" and is what you see when you type `git log --oneline`

parkmichael777: Okay, changed and pushed!

bthayer2365:
fixes #91

gitgud/operations.py
parkmichael777: Done

parkmichael777:
Done! I didn't think to do list slicing because I didn't know if the length of the message is different every time lol. Changed, and added goal stuff.

bthayer2365:
I modified the goal, but otherwise, things looked good, and I merged!

PR #67, opened 2019-10-23
Added detaching instructions, test, and setup

parkmichael777:


bthayer2365:
From #24:
#### 1.

I don't think detaching head is quite clear what to do. Can you add more specific instructions? We don't quite have the benefit of displaying the goal state yet, so we'll need to have more specific instructions.

parkmichael777:
Added the end goal to the instructions.

gitgud/levels/rampup/__init__.py
```diff
@@ -0,0 +1,11 @@
+from collections import OrderedDict
+
+import pkg_resources
+
+from gitgud.levels.util import BasicChallenge
+
+all_challenges = OrderedDict()
+all_challenges['detaching'] = BasicChallenge('detaching', pkg_resources.resource_filename(__name__, '_detaching/'))
```
bthayer2365: Can you remove all challenges except `detaching`?

bthayer2365:
Everything looks good, but we shouldn't add the other challenges until they're finished

parkmichael777:
Should be removed for all four PRs

bthayer2365:
Merged!

PR #68, opened 2019-10-23
Added relrefs1 instructions, setup, and test

parkmichael777:


bthayer2365:
From #24:

#### 2.

We didn't learn that we only need 4 digits of the commit hash.

I'd also call it "another way" instead of a "more efficient way". Relative refs are used when it's just a few commits back, but for example, when I'm looking though the git logs to see the commit message, I'd rather not take the time to count back how many commits that is when I can read that the hash is 8636.... In that case, reading the hash is more efficient.

^ and ~ are not quite how you explained them. ^ is shorthand for ^1, and ~ can be used as shorthand for ~1. ^ is used to say which parent to go to. ^1 goes to parent 1, which in the case of a normal commit, is just the parent. ^2 will go to the merged commit. If there are more parents, like in an octopus merge, then ^3 and so on can be used, but there's no need to mention that in this level. Maybe in a future level, but for now, I think it's safe to keep it out. Similarly, ~ and ^ can be stringed together. ~~~ is possible, just like ^^^ is possible. Similarly, different sequences can be strung together like ~^~ or ~^2~5, which would get the merge commit that happened prior to the current commit, access the other commit that was merged into the original branch and go five back from there.

Also, unfortunately, on windows (but not on git bash), ^ is the character that allows you to let a command take up mulitple lines. Instead, on windows we use ^^, which escapes the character and converts it to an actual ^. This should be noted.

The instructions explain stuff about relative refs but don't explain what to do to solve the level, so can you add that?

parkmichael777:
In the instructions of detaching, I explain that you only need 4 digits of the hash to refer to the commit. Regardless, I explain it again when I mention it in the relrefs1 instructions. 

Changed to "more efficient ways"

Added explicit explanation of the ^ operator referring to parent.

Added windows-specific note

Added explicit end goal sentence.

gitgud/levels/rampup/__init__.py
```diff
@@ -0,0 +1,11 @@
+from collections import OrderedDict
+
+import pkg_resources
+
+from gitgud.levels.util import BasicChallenge
+
+all_challenges = OrderedDict()
+all_challenges['detaching'] = BasicChallenge('detaching', pkg_resources.resource_filename(__name__, '_detaching/'))
+all_challenges['relrefs1'] = BasicChallenge('relrefs1', pkg_resources.resource_filename(__name__, '_relrefs1/'))
```
bthayer2365: Can you remove all challenges except `relrefs1`?

gitgud/levels/rampup/_relrefs1/instructions.txt
```diff
@@ -0,0 +1,15 @@
+Now that we've learned how to detach the HEAD, let's learn easier ways to refer to a commit.
+>>>
+In the previous exercise, we learned that we only need 4-digits of a commit hash in order to reference a particular commit. This method made the process much easier than entering in a giant string of characters and numbers (e.g. fed2da64c0efc5293610bdd892f82a58e8cbc5d8). However, in this exercise we'll learn even more ways of refering to commits!
+>>>
+In particular there are two ways of referencing commits called "relative refs." Relative refs are used to refer to a commit nodes that are a couple commits behind your current commit. It should be noted that using relative refs are only efficient to use if the commit you are referring is nearby in the tree; otherwise, in many cases it may be faster to bring up the commit log and use the hash.
```
bthayer2365: Can you call them "relative references" when you introduce them just to be clear? After `called "relative references"`, I think it's fine to call them refs, people should be able to pick up on that

gitgud/levels/rampup/_relrefs1/instructions.txt
```diff
@@ -0,0 +1,15 @@
...
+>>>
+The ^ operator refers to the parent node of the commit you specify. I.e., "git checkout bugFix^" will checkout the parent node of bugFix! "^" is shorthand for "^1" meaning parent 1. To further illustrate this point, say commit "C" is the merge node of "A" and "B." "C^2" refers to parent two of C, which is B. For an octopus merge, "^3" would refer to parent 3, and so on. The second operator is the "~<number>" operator. It's used to point to a commit that is several nodes above the node you specify. That operator will be covered in the next exercise. For now, let's take a look at the caret operator.
```
bthayer2365: `is several nodes above the node you specify` -> `is several commits prior to the node you node you specify`

parkmichael777:
Applied those changes

bthayer2365:
Looks good! I added one commit to correct the change I told you to make and say commit instead of node, but it's all good, and you're merged!

Issue #34, opened 2019-10-18
Assign tags to commits based on their names


benthayer:
Might be unnecessary, see #87 

In spec files, we name each of our commits, and it would be nice to be able to check them out using `git checkout C1`

Commit tags should be a capital "C" followed with the number, no spaces. Merge commits should follow the same pattern but with an "M", just as it looks in the spec files.

We'll need to make modifications to `create_tree` and `add_and_commit` in `gitgud/operations.py`

PR #69, opened 2019-10-23
Added relrefs 2 instructions, test, setup; Modified operations to check for detached head

parkmichael777:


bthayer2365:
From #24:

#### 3.

setup.spec and level.spec are a little wacky and don't really correspond to the instructions, please change those. Either clarify in the instructions what you mean or change the specs to align with the instructions more.

This also works with branches -> this also works with branches and tags, and will work any other reference

Again, instead of saying X commits "above", use the language I described above

I think you can remove this whole line: A useful application of the tilde operator is the command "git branch -f <branch_name> HEAD~<num>". This command allows you to move a certain branch to a certain commit node <num> above the current node. That isn't the most common use of this, but it certainly can be done and we should teach them how to do that, but still one thing at a time. You should open up a new issue to make a new level based on git branch -f <branch_name> HEAD~. Another challenge that might go in the same level as that is using the git reset --hard HEAD~, which would modify the current branch. Add those both to the issue. For future reference, "above" is an ambiguous term, you should say <num> commits into the history/in the past or the previous <num>th commit

bthayer2365:
This PR should also handle #34

parkmichael777:
Added new instructions for a new challenge. The old challenge used the "git branch -f <branch_name> HEAD~" commands, but that has been removed for this challenge.

This is already mentioned in the instructions.

Changed language from "<num> above" to "<num> in the past."

Removed this line.

gitgud/levels/rampup/__init__.py
```diff
@@ -0,0 +1,11 @@
+from collections import OrderedDict
+
+import pkg_resources
+
+from gitgud.levels.util import BasicChallenge
+
+all_challenges = OrderedDict()
+all_challenges['detaching'] = BasicChallenge('detaching', pkg_resources.resource_filename(__name__, '_detaching/'))
+all_challenges['relrefs1'] = BasicChallenge('relrefs1', pkg_resources.resource_filename(__name__, '_relrefs1/'))
+all_challenges['relrefs2'] = BasicChallenge('relrefs2', pkg_resources.resource_filename(__name__, '_relrefs2/'))
```
bthayer2365: Can you remove all challenges except `relrefs2`?

gitgud/levels/rampup/_relrefs2/instructions.txt
```diff
@@ -0,0 +1,9 @@
+Now that you are a master with the ^ operator, let's learn how to use the ~ operator.
+>>>
+The tilde operator is used much the same way as the ^ operator; although it can be more concise than the ^ operator. The usage of the ~ operator is this: "git checkout HEAD~<num>".
+>>>
+Using this command will place the HEAD pointer at the commit node that is <num> commits in the past. So, if I want to check out the node 5 commits before HEAD, I would use: "git checkout HEAD~5". This command also works with branch names instead of HEAD.
```
bthayer2365: Just like ^, this command also works with other types of references like branch names, commit hashes and tags, instead of just HEAD

bthayer2365:
Just copy what I said in the one note and we'll be good for a merge. Challenge works as expected

parkmichael777:
Added that to the instructions

bthayer2365:
Looks good, thanks!

Issue #26, opened 2019-10-18
Allow loading by level/challenge numbers


benthayer:
Right now, we just load challenges by their names, but it might be useful to load them by their challenge numbers in addition to by name.

This issue can't be fixed until #23 is complete.

bthayer2365:
@sahansk2 Would you also like to handle this?

bthayer2365:
You'll need to make sure to show the numbers when running `git gud skills`

sahansk2:
Sure, I'll take on this issue. I'll open a pull request for it as well.

PR #70, opened 2019-10-24
Data structure to access levels and challenges by index

### LINKED ISSUE #23: Change the data structure for all_levels and all_challenges to allow indexing

sahansk2:
fixes #23.
Nothing implemented yet, but here's a PoC for how it will be implemented:
```
class NamedList:
    # feed in tuple (reference, Object), where the reference is of type str
    def __init__(self, data=[]):
        # Pass in a list with tuples
        self.namehash = {data[i][0]:i for i in range(len(data))}
        self.indexedlist = [obj for nam,obj in data]
    def __getitem__(self, query):
        if type(query) is int:
            return self.indexedlist[query]
        elif type(query) is str:
            return self.indexedlist[self.namehash[query]]
        else:
            return None

# # Sample code to test functionality of the data structure. 
# class Animal:
#     def __init__(self, name, kind, cry):
#         self.name = name
#         self.cry = cry
#         self.kind = kind
#     def callout(self):
#         print("My name is {}! I am a {} and I say {}!".format(self.name, self.kind, self.cry))

# d = Animal('Foo', 'dog', 'arf')
# c = Animal('Bar', 'cat', 'meow')

# nl = NamedList([
#     ('Foo', d), 
#     ('Bar', c)
# ])

# nl['Foo'].callout()
# nl[1].callout()
```


sahansk2:
As many instances of values() being called have been removed as well, to improve readability. However, I left the values() function in the NamedList class definition, if desired in the future.

I did not make the Challenge and Level classes inherit from NamedList, because I still need to understand more on how those classes work in order to safely let them inherit from NamedList. But, this shallower refactoring works for the time being, and we can work on other issues that are dependent on this issue (like #26 ). What do you think, @bthayer2365 ?

bthayer2365:
I don't think we need to stop where you've left it, you've done most of the work, now we have to plug in you class and put it to use. Inheritance has a big name, but it's really not all that complex. You just specify `class SubClass(SuperClass):` and `SubClass` will be able to run all the methods in `SuperClass` by calling `self.superclassmethod()`. It'll also have access to other attributes. For example, a subclass of NamedList will have `self.namedict` and `self.items`.

I think it makes sense to go bit by bit here, so the first thing is to subclass NamedList for both AllLevels and Level. Once you do that, we should automatically be able to do things like `for level in all_levels:` and `all_levels['intro']['committing']`

In the subclasses, we can then use the initializers to set `namedict` for us. I'll add some more comments in the code, but to start:
1. Subclass NamedList 
2. Use the init method to set the values of `namedict`

gitgud/__main__.py
```diff
@@ -132,8 +132,8 @@ def handle_start(self, args):
         with open(self.file_operator.last_commit_path, 'w+') as commit_file:
             commit_file.write('0')  # First commit will be 1
         with open(self.file_operator.level_path, 'w+') as level_file:
-            level1 = next(iter(all_levels.values()))
-            challenge1 = next(iter(level1.challenges.values()))
+            level1 = all_levels[0]
+            challenge1 = level1.challenges[0]
             level_file.write(challenge1.full_name())
```
bthayer2365: We can change this into one line
```
level.write(all_levels[0][0].full_name())
```

gitgud/__main__.py
```diff
@@ -197,11 +197,11 @@ def handle_levels(self, args):
 
         print("Currently on level: \"{}\"\n".format(cur_level.name))
         
-        for level in all_levels.values():
+        for level in all_levels:
             # TODO Add description
             # 10 characters for the short IDs. 
             print("Level {:<10} :{:>2} challenge{}".format("\"" + level.name + "\"", len(level.challenges), ("", "s")[len(level.challenges) > 1]))
-            for index, challenge in enumerate(level.challenges.values()):
+            for index, challenge in enumerate(level.challenges):
```
bthayer2365: Once we update, we should be able to do `for index, challenge in enumerate(level):`

gitgud/__main__.py
```diff
@@ -224,7 +224,7 @@ def handle_challenges(self, args):
             print("Challenges for level \"{}\" : \n".format(level.name))
 
         
-        for index, challenge in enumerate(level.challenges.values()):
+        for index, challenge in enumerate(level.challenges):
```
bthayer2365: Same here,

gitgud/__main__.py
```diff
@@ -241,7 +241,7 @@ def handle_load(self, args):
                     print("Challenge \"{}\" does not exist".format(args.challenge_name))
                     print("To view challenges/levels, use git gud challenges or git gud levels")
             else:
-                challenge = next(iter(level.challenges.values()))
+                challenge = next(iter(level.challenges))
                 challenge.setup(self.file_operator)
```
bthayer2365: We can make this one line similar to above: `level[0].setup(self.file_operator)`

gitgud/levels/__init__.py
```diff
@@ -1,20 +1,18 @@
-from collections import OrderedDict
-
 from gitgud.levels.intro import all_challenges as intro_challenges
 from gitgud.levels.merging import all_challenges as merging_challenges
 
 from gitgud.levels.util import Level
+from gitgud.levels.util import NamedList
```
bthayer2365: all_levels should be of type AllLevel, we won't nee this import

gitgud/levels/__init__.py
```diff
@@ -1,20 +1,18 @@
...
 
-all_levels = OrderedDict()
-
-all_levels['intro'] = Level('intro', intro_challenges)
-all_levels['merging'] = Level('merging', merging_challenges)
-
+all_levels = NamedList([
+    ('intro', Level('intro', intro_challenges)),
+    ('merging', Level('merging', merging_challenges))
+])
```
bthayer2365: AllLevels should use `__init__` to grab the name from each level, so specifying the names in the tuples would be unneccessary.

gitgud/levels/__init__.py
```diff
@@ -1,20 +1,18 @@
...
 
 def _add_next_challenges():
```
bthayer2365: I think we can move this to the constructor for AllLevels

gitgud/levels/intro/__init__.py
```diff
@@ -1,11 +1,11 @@
...
 
 from gitgud.levels.util import BasicChallenge
+from gitgud.levels.util import NamedList
 
-all_challenges = OrderedDict()
-all_challenges['committing'] = BasicChallenge('committing', pkg_resources.resource_filename(__name__, '_committing/'))
-all_challenges['branching'] = BasicChallenge('branching', pkg_resources.resource_filename(__name__, '_branching/'))
-all_challenges['merging'] = BasicChallenge('merging', pkg_resources.resource_filename(__name__, '_merging/'))
-all_challenges['rebasing'] = BasicChallenge('rebasing', pkg_resources.resource_filename(__name__, '_rebasing/'))
+all_challenges = NamedList([
```
bthayer2365: Make this of type Level, and once again, forego specifying the names in the tuple and instead use the constructor of level to handle the names

gitgud/levels/__init__.py
```diff
@@ -1,20 +1,18 @@
-from collections import OrderedDict
-
 from gitgud.levels.intro import all_challenges as intro_challenges
 from gitgud.levels.merging import all_challenges as merging_challenges
 
 from gitgud.levels.util import Level
```
bthayer2365: You won't need to import this because you'll be importing the levels from the subpackages above

gitgud/levels/intro/__init__.py
bthayer2365: Rename this to level since that's what the new version will be

gitgud/levels/merging/__init__.py
```diff
@@ -1,8 +1,8 @@
...
 
 from gitgud.levels.util import BasicChallenge
+from gitgud.levels.util import NamedList
 
-all_challenges = OrderedDict()
-all_challenges['octopus'] = BasicChallenge('octopus', pkg_resources.resource_filename(__name__, '_octopus/'))
\ No newline at end of file
+all_challenges = NamedList([
+    ('octopus', BasicChallenge('octopus', pkg_resources.resource_filename(__name__, '_octopus/')))
+])
```
bthayer2365: Comments for level intro apply here too

gitgud/levels/util.py
```diff
@@ -194,12 +192,40 @@ def test_level(level, test):
 
     return True
 
-
-class Level:
+class NamedList:
+    # Pass in list of tuples in format (reference, Object), where reference is of type str
+    def __init__(self, data=[]):
+        self.namedict = {d[0]:i for i, d in enumerate(data)}
```
bthayer2365: Rename to `self._name_dict`. First underscore indicates that the attribute is private, second underscore is convention and makes it easier to read

gitgud/levels/util.py
```diff
@@ -194,12 +192,40 @@ def test_level(level, test):
...
+        self.items = [obj for name, obj in data]
```
bthayer2365: Rename to `self._items`, again because it should be private

gitgud/levels/util.py
```diff
@@ -194,12 +192,40 @@ def test_level(level, test):
 
     return True
 
-
-class Level:
+class NamedList:
+    # Pass in list of tuples in format (reference, Object), where reference is of type str
+    def __init__(self, data=[]):
```
bthayer2365: I think it would be better to make the definition `def __init__(self, names, items):`, it just makes things more clear. We'll need to assert that they're the same length `assert len(names) == len(items)`, and then we can use `for name, item in zip(names, items)`

gitgud/levels/util.py
```diff
@@ -194,12 +192,40 @@ def test_level(level, test):
...
+            self.items.append(item)
+        else:
+            raise TypeError
+    
+    def values(self):
+        return self.items
+    def keys(self):
+        return self.namedict.keys()
+
+class Level():
```
bthayer2365: Make a new class AllLevels and make NamedList a superclass of both Level and AllLevels

gitgud/levels/util.py
```diff
@@ -194,12 +192,40 @@ def test_level(level, test):
...
     def __init__(self, name, challenges):
         self.name = name
-        self.challenges = OrderedDict()
-        for challenge in challenges.values():
+        self.challenges = NamedList()
```
bthayer2365: When Level is a subclass of NamedList, we can use `self._name_dict` and `self._items` instead of specifying a list of challenges

bthayer2365:
If monsters weren't scary enough, how about this review? Boo! Happy Halloween.. haha. Make as much progress as you can. Go in whatever order you like and comment on each of my reviews as you finish them and I'll be able to review as you go. If you need help or clarification, just ask

sahansk2:
I've done everything you've said thus far! But, in just testing out the code, some of the parser commands simply don't work anymore (for example: `git gud load intro` generates an output of `Level "intro" not found` and `git gud challenges` exits with an error). I've pushed what I have so far, in case you would also like to skim through to see what must be changed.

sahansk2:
All `git gud` commands seem to be in proper shape -- the problem was that `handle_load()` was searching through the list of `Level` objects for a `str`, not through the actual _names_ of the `Level` objects. 
All cases where a Level object uses challenges have been removed. 
Everything should be good now!

gitgud/__main__.py
```diff
@@ -224,26 +223,25 @@ def handle_challenges(self, args):
...
 
         
-        for index, challenge in enumerate(level.challenges):
+        for index, challenge in enumerate(level):
             print(str(index + 1) + ": " + challenge.name)
 
     def handle_load(self, args):
         self.assert_initialized()
-        if args.level_name in all_levels:
+        if args.level_name in all_levels.keys():
```
bthayer2365: You should just be able to check for the level name in the all_levels object, you don't need to access the keys

gitgud/__main__.py
```diff
@@ -224,26 +223,25 @@ def handle_challenges(self, args):
...
             level = all_levels[args.level_name]
 
             if args.challenge_name is not None:
-                if args.challenge_name in all_levels[args.level_name].challenges:
-                    challenge = level.challenges[args.challenge_name]
+                if args.challenge_name in all_levels[args.level_name].keys():
```
bthayer2365: Same as all_levels, you should be able to use what you had before. If not, there is a `__contains__` method you can implement

gitgud/__main__.py
bthayer2365: You might want the `__contains__` method, but I'm not sure. You can test that

gitgud/levels/intro/__init__.py
```diff
@@ -1,11 +1,14 @@
...
+from gitgud.levels.util import Level
 
-all_challenges = NamedList([
-    ('committing', BasicChallenge('committing', pkg_resources.resource_filename(__name__, '_committing/'))),
-    ('branching', BasicChallenge('branching', pkg_resources.resource_filename(__name__, '_branching/'))),
-    ('merging', BasicChallenge('merging', pkg_resources.resource_filename(__name__, '_merging/'))),
-    ('rebasing', BasicChallenge('rebasing', pkg_resources.resource_filename(__name__, '_rebasing/')))
-])
+all_challenges = Level(
+    'level',
```
bthayer2365: level name should be 'intro'

gitgud/levels/intro/__init__.py
```diff
@@ -1,11 +1,14 @@
...
-from gitgud.levels.util import NamedList
+from gitgud.levels.util import Level
 
-all_challenges = NamedList([
-    ('committing', BasicChallenge('committing', pkg_resources.resource_filename(__name__, '_committing/'))),
-    ('branching', BasicChallenge('branching', pkg_resources.resource_filename(__name__, '_branching/'))),
-    ('merging', BasicChallenge('merging', pkg_resources.resource_filename(__name__, '_merging/'))),
-    ('rebasing', BasicChallenge('rebasing', pkg_resources.resource_filename(__name__, '_rebasing/')))
-])
+all_challenges = Level(
```
bthayer2365: you can call this level instead of all_challenges

gitgud/levels/__init__.py
```diff
@@ -2,20 +2,21 @@
 from gitgud.levels.merging import all_challenges as merging_challenges
```
bthayer2365: You can change this line to `from gitgud.levels.merging import level as merging_level`. You'll also need to change `all_challenges` to `level` for this line to work. Do the same thing for the merging level

gitgud/levels/__init__.py
```diff
@@ -2,20 +2,21 @@
...
+from gitgud.levels.util import AllLevels
 
-all_levels = NamedList([
-    ('intro', Level('intro', intro_challenges)),
-    ('merging', Level('merging', merging_challenges))
-])
+all_levels = AllLevels(
+    [
+        Level('intro', intro_challenges),
+        Level('merging', merging_challenges)
```
bthayer2365: replace this with `intro_level` and `merging_level` as suggested above

gitgud/levels/merging/__init__.py
```diff
@@ -1,8 +1,11 @@
...
 from gitgud.levels.util import BasicChallenge
-from gitgud.levels.util import NamedList
+from gitgud.levels.util import Level
 
-all_challenges = NamedList([
-    ('octopus', BasicChallenge('octopus', pkg_resources.resource_filename(__name__, '_octopus/')))
-])
\ No newline at end of file
+all_challenges = Level(
+    'level',
```
bthayer2365: Level name should be 'merging'

gitgud/levels/merging/__init__.py
```diff
@@ -1,8 +1,11 @@
...
 
 from gitgud.levels.util import BasicChallenge
-from gitgud.levels.util import NamedList
+from gitgud.levels.util import Level
 
-all_challenges = NamedList([
-    ('octopus', BasicChallenge('octopus', pkg_resources.resource_filename(__name__, '_octopus/')))
-])
\ No newline at end of file
+all_challenges = Level(
```
bthayer2365: `level` instead of `all_challenges`

gitgud/levels/util.py
```diff
@@ -193,42 +193,51 @@ def test_level(level, test):
...
 class NamedList:
-    # Pass in list of tuples in format (reference, Object), where reference is of type str
-    def __init__(self, data=[]):
-        self.namedict = {d[0]:i for i, d in enumerate(data)}
-        self.items = [obj for name, obj in data]
+    # names is a list populated with type str, items is a list populated with any type 
+    def __init__(self, names, items):
+        assert len(names) == len(items)
+        self._name_dict = {name:index for index, name in enumerate(names)}
+        self._items = [item for name, item in zip(names, items)]
```
bthayer2365: You can do `self._items = items`

gitgud/levels/util.py
```diff
@@ -193,42 +193,51 @@ def test_level(level, test):
...
     
     def __getitem__(self, query):
         if isinstance(query, int):
-            return self.items[query]
+            return self._items[query]
         elif isinstance(query, str):
-            return self.items[self.namedict[query]]
+            return self._items[self._name_dict[query]]
         else:
             return None
```
bthayer2365: You should change this to `raise ValueError('Level not found')`. When things fail, we should raise an exception unless we for some reason intend on using that `None` value

gitgud/levels/util.py
```diff
@@ -193,42 +193,51 @@ def test_level(level, test):
...
 
-class Level():
+class Level(NamedList):
     def __init__(self, name, challenges):
         self.name = name
-        self.challenges = NamedList()
-        for challenge in challenges:
+        self._name_dict = {challenge.name:index for index, challenge in enumerate(challenges)}
+        self._items = challenges
+        for challenge in self._items:
```
bthayer2365: I think we can change `self._items` to `challenges` for clarity. Also add a blank line before the for loop just to add some visual separation between setting the variables and then processing them

gitgud/levels/intro/__init__.py
bthayer2365: Resolving this thread becuase I commented again on the new version

bthayer2365:
Just went back through and resolved the comments that you fixed. Everything you see should be the changes you have to make. I still haven't tested. I'll do that once the code looks good.

gitgud/levels/__init__.py
```diff
@@ -1,23 +1,22 @@
...
-
-all_levels['intro'] = Level('intro', intro_challenges)
-all_levels['merging'] = Level('merging', merging_challenges)
-
+all_levels = AllLevels(
+    [
+        intro_level,
+        merging_level
+    ]
+)
```
bthayer2365: Looks, good. As a style preference, let's put the ([ and ]) on the same line and change the indentation, so 
```
all_levels = AllLevels([
    intro_level,
    merging_level
])
```

gitgud/levels/util.py
```diff
@@ -231,14 +237,21 @@ def __init__(self, name, challenges):
...
     def __init__(self, levels):
         self._name_dict = {level.name:index for index, level in enumerate(levels)}
         self._items = levels
 
+    def __contains__(self, key):
+        if isinstance(key, str):
+            return key in self._name_dict.keys()
+        else:
+            return key in self._items()
+
```
bthayer2365: You should be able to delete this method. It'll be inhereted from NamedList

bthayer2365:
Scroll through the history of this conversation. There are a few things that still need to get done, some from before, some from your new updates. Just a few though

gitgud/levels/util.py
```diff
@@ -194,15 +192,67 @@ def test_level(level, test):
...
+        for challenge in challenges:
             challenge.level = self
-            self.challenges[challenge.name] = challenge
 
+class AllLevels(NamedList):
+    def __init__(self, levels):
+        self._name_dict = {level.name:index for index, level in enumerate(levels)}
+        self._items = levels
+    
+    def _add_next_challenges(self):
```
bthayer2365: You can remove this method definition entirely and just put the code in the constructor since we'll want it to be called every time no matter what

gitgud/levels/__init__.py
```diff
@@ -1,23 +1,11 @@
...
-    last_challenge = None
-    for level in all_levels.values():
-        for challenge in level.challenges.values():
-            if last_challenge is not None:
-                last_challenge.next_challenge = challenge
-            last_challenge = challenge
-
-
-_add_next_challenges()
+all_levels._add_next_challenges()
```
bthayer2365: You can delete this line. The constructor will handle this

gitgud/levels/__init__.py
```diff
@@ -1,22 +1,11 @@
 from gitgud.levels.intro import level as intro_level
 from gitgud.levels.merging import level as merging_level
 
-from gitgud.levels.util import Level
 from gitgud.levels.util import AllLevels
 
-all_levels = AllLevels(
-    [
+all_levels = AllLevels([
         intro_level,
```
bthayer2365: Can you reduce this to one tab (4 spaces)

bthayer2365:
We did it! You're merged. Up next is #26

Issue #62, opened 2019-10-23
Log commit hashes when creating commits


bthayer2365:
This is applicable to `git gud commit` and calls to `create_tree()`. This should be done inside the  `create_tree` and inside the `handle_commit` method.

This is useful because it gives the user information about the events that are happening to create the current tree.

Edit: logging for `git gud commit` is complete due to #71. We still need to do this for `create_tree()`

cheedep:
@bthayer2365 how should the messages look like. Is it possible to get an example for each of `git gud commit` and `create_tree()`

For example this is what I have for `git gud commit`
![image](https://user-images.githubusercontent.com/526575/67453858-ee3b8880-f5dc-11e9-861c-cacdcab11da5.png)

I am yet to figure out the create_tree()

bthayer2365:
What you have there is good, thanks for working on this.

I pasted below what it normally looks like when committing. 
```
[master f7778e8] Added 3
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 3
```

I think we can cut down the hash to 8 characters like git does normally and maybe change the message you have to be prettier.

So, something like 
```
...
Simulating: git commit -m "3"
New Commit: d77413e8
```

Also, I took a look at your code, and unfortunately, for comparability reasons, we can't use f-strings 😢 but oh I do love them so much! f-strings were introduced in python 3.6, and some people only have 3.5 or lower, so to make it easier for everyone to install, we need to avoid using f-strings in favor of `'string'.format()`

PR #71, opened 2019-10-24
log commit hash after git gud commit

cheedep:
partial implementation for #62


bthayer2365:
Hi @cheedep , are you still working on this? 

Here's what still needs to be done:
You already show the commit hash when running `handle_commit`, so that's good.
String format is safe for Python<3.6, so that's good too.

There's still a couple more things:
Can you show the first 8 characters instead of 7?
Can you copy the print statements from handle_commit into create_tree?

cheedep:
@bthayer2365 I will work on it today. for some reason I thought github showed 8 characters for the commit hash but I see it is only 7.

bthayer2365:
Hi @cheedep, I merged your code. If you'd like to add the same feature for create_tree, go ahead and make a new PR. Thanks for contributing!

Issue #63, opened 2019-10-23
Automatically show tree when loading level


benthayer:
This would be the same as calling `git gud show-tree` right after `git gud load X` or `git gud reset`. Anything that loads the level

AumkarR:
Hey, I'm going to start working on it

PR #72, opened 2019-10-24
automatically show tree after load and reset

cheedep:
#63

### DIFF
--- a/gitgud/__main__.py
+++ b/gitgud/__main__.py
@@ -166,6 +166,7 @@ class GitGud:
         challenge = self.file_operator.get_challenge()
         print("Resetting...")
         challenge.setup(self.file_operator)
+        self.handle_show_tree(args)
 
     def handle_test(self, args):
         self.assert_initialized()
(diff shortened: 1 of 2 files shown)


gitgud/__main__.py
```diff
@@ -185,13 +187,17 @@ def handle_progress(self, args):
...
         if next_challenge is not None:
-            next_challenge.setup(self.file_operator)
-            self.file_operator.write_challenge(next_challenge)
+            self.handle_load_challenge(args, next_challenge)
         else:
             print("Wow! You've complete every challenge, congratulations!")
             print("If you want to keep learning git, why not try contributing to git-gud by forking us at https://github.com/bthayer2365/git-gud/")
             print("We're always looking for contributions and are more than happy to accept both pull requests and suggestions!")
 
+    def handle_load_challenge(self, args, challenge):
```
bthayer2365: `handle_*` methods should be reserved for the subcommands, so we can just rename this to `load_challenge`. Since it's not a handler anymore, it doesn't have to accept `args` anymore.  and have it accept just one argument, `challenge`.

bthayer2365:
Thanks for working on this! Other than the changes I suggested, I think everything looks good!

bthayer2365:
Looks good! Thanks for adding this

Issue #77, opened 2019-10-27
Hooks being ignored due to them not being marked as executable


sahansk2:
**Describe the bug**
Git hooks do not run when their corresponding git commands are executed in a git-gud directory.

**To Reproduce**
Run `git commit --amend` in an initialized git-gud directory to test the `post-rewrite` hook.

I have run `git gud start` in multiple fresh directories to see if it was a problem with directories already initialized and manipulated with other git commands, but it appears in the fresh directories as well.

**Expected behavior**
The git hooks should run without any problem, and without the user running the suggested command from git to ignore the corresponding warning.

**Screenshots/Output**
`$ git commit --amend`
```
hint: The '.git/hooks/post-rewrite' hook was ignored because it's not set as executable.
hint: You can disable this warning with `git config advice.ignoredHook false`.
[master 194c4cd] first commit
 Author: Git Gud <git-gud@example.com>
 Date: Sat Oct 26 21:20:50 2019 -0500
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 1
```
**Environment**
OS: Ubuntu 18.04.3 LTS, 64-bit
python Version: 3.6.8
git Version: 2.17.1

divCurlF:
Fixed in #79

PR #78, opened 2019-10-27
Allows a variable number of arguments to be passed into git hooks. 

### LINKED ISSUE #45: Allow git hooks to accept a varying amount of arguments
Right now, git hooks only accept one argument. That's because I've only tested with the `post-rewrite` hook, which accepts one argument. It's unclear whether hooks with a different number of arguments actually work right now. To solve this issue, you should test it and see. 

You'll need to do some work with bash to solve this issue since the script that calls the git hooks is written in bash. 

To see the created file, just run `git gud start` and see the created `post-rewrite` file in the `.git/hooks` directory. If I were solving this issue, I would make a script that looks similar and a python script to work with it and see what you get, but do it however you'd like.

sahansk2:
Fixes #45.

This has been tested with a custom script using the git hook `prepare-commit-msg`, which takes up to three arguments. This test script has not been included in the pull request.

Not taking #77 into consideration, both the custom `prepare-commit-msg` python script and the already existing `post-rewrite` python script work with git hooks, despite the fact that they take different numbers of inputs.

(As an aside: my commits from the very beginning of my involvement are appearing in this PR thread, is this normal?)

### DIFF
--- a/gitgud/__main__.py
+++ b/gitgud/__main__.py
@@ -280,12 +280,12 @@ class GitGud:
         subprocess.call(["git", "log", "--graph", "--oneline", "--all"])
 
     def parse(self):
-        args = self.parser.parse_known_args()
-        if args[0].command is None:
+        args, badargs = self.parser.parse_known_args()
+        if args.command is None:
             self.parser.print_help()
         else:
             try:
-                self.command_dict[args[0].command](args[0])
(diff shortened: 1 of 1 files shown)


bthayer2365:
Perfect, thanks. I think this was the easiest review I've ever done 👍

PR #76, opened 2019-10-27
Added level "cherrypicking"

JakeDuo:
added challenge cherrypicking

Issue #43, opened 2019-10-19
Complete the `post-rewrite` hook to keep track which commits have been rebased


benthayer:
We'll need to put this information in the json file created by #40. When a commit is rebased, we'll want to remember the hash of the previous commit, and we'll want to remember the new commit so we can check if the commit was actually rebased. Once this is implemented, we'll want to update the levels that involve rebasing to take advantage of the newly defined git hook.

Learn Git Branching represents rebased commits with an apostrophe, such that after a rebase, `1` would become `1'`. It's interesting to note that the command line doesn't get rid of the single quote while the double-quote is treated like a special character and isn't allowed to be used in a commit name.

If rebasing a rebased commit, add another `'`, so `1'` rebased is `1''`

bthayer2365:
@jakeduo, would you like to work on this?

JakeDuo:
Sure, I'll ask you if I need any help figuring things out.

benthayer:
With the addition of #104, we'll need to make sure we add the rebase notation to the spec file templates

benthayer:
@JakeDuo How's this coming? You can work with @parkmichael777 on this and #87. If either of you need help, let me know

benthayer:
@JakeDuo 

I'm trying to get Git Gud into beta by the beginning of break so I can start talking to people and seeing how they interact with it and then add some more final features over break and hopefully start getting professors to use it in their classes next semester. Can you get together within the next week or so, so I can help you get started on this?

parkmichael777:
@benthayer @JakeDuo This is next on the todo list. I can do it in a couple lines using the stuff already there and the json tracking code I wrote. Can I take this issue?

JakeDuo:
Sure

JakeDuo:
@parkmichael777 Sure

parkmichael777:
Fixed by #132

PR #84, opened 2019-10-31
Added level "One Commit"

JakeDuo:
one commit level under the mixed bag section


bthayer2365:
Can you rebase this so you only have the 36f37c2535e2e3c79b07d7810be369b0e9a85b19 commit? 

You can rebase off of 42b87f4b1075acd5aa70756fa1ed95ff91cf7a6d

bthayer2365:

gitgud/levels/mixedbag/_onecommit/setup.spec
```diff
@@ -0,0 +1,5 @@
+1 (master)
+2 : 1 (debug)
```
bthayer2365: This would be the perfect opportunity to use tags instead of commits. To specify a tag, do `(tag: debug)`. Do this for debug/printf for both spec files

gitgud/levels/mixedbag/_onecommit/test.spec
```diff
@@ -0,0 +1,6 @@
+1
+2 : 1 (debug)
```
bthayer2365: `2:1` can be `2`. Can you change this for both files?

bthayer2365:
#43 will allow us to test for rewriting

benthayer:
So, we've still got the changes I requested from before, and then you need to add the test cases like in #76, but this time with tricks instead of rework

PR #85, opened 2019-11-01
Added level "Juggling Commits 1"

JakeDuo:
level for jugglecommits1 in mixed bag section


bthayer2365:
Can you rebase this so you only have the 574cde5719c98925c86680526430c5bd0d27651b commit? 

You can rebase off of 42b87f4b1075acd5aa70756fa1ed95ff91cf7a6d

bthayer2365:
@JakeDuo would you like me to revert/squash these changes for you?

JakeDuo:
@bthayer2365 Yea, I've been busy lately.

bthayer2365:

bthayer2365:
I'll just comment on this one to explain what I did for all your PRs. The commits I added revert all of your work then re-applies the work that is specific to each pull request. Once I review them and you get all the changes done, I'll rebase and get rid of all the initial commits, keep the commits after all the reverting, then I'll force push so the commit history is clean, and then I'll merge.

gitgud/levels/mixedbag/_jugglecommits1/test.spec
```diff
@@ -0,0 +1,9 @@
+1
+2 : 1 (newImage)
+3
+3' : 1
+2' : 3'
+2'' : 3'
+2''' : 1
+c3'' (caption)(master)
```
bthayer2365: The syntax here should be `3'' (caption, master)`

bthayer2365:
Also, tags are necessary because otherwise we can't find those commits

bthayer2365:
#43 will allow us to test for rewriting

benthayer:
> 
> 
> Thanks for doing this. This level is a little tricky because we don't have the benefit of the visualization tools of LGB. I think instead of telling them what to do, they're supposed to figure it out. Maybe it's best to just show them a commit tree like this and tell them to find a way to copy it:
> 
> ```
> * 1da9d97 (HEAD -> master, caption) 3''
> * c06bcee 2'''
> | * 2cf3676 (tag: two) 2''
> | | * 4b43c84 (tag: one) 2'
> | |/
> | * dcf2fdb 3'
> |/
> | * 8550051 (tag: start) 3
> | * 949ab67 (newImage) 2
> |/
> * 0b38f4a 1
> ```
> 
> We could also modify that to be diagonal or something.
> 
> What do you think?

Can you respond to this?

Same with the others, you have to include the test cases

benthayer:
With the new updates, there's a bit of an issue with the syntax. Our current commit processor only accepts single rebases, not doubles. We'll need to figure out whether we want to implement that or find another way to get through the level

benthayer:
@sahansk2 given that we're removing levels from learngitbranching, what do you think we should do with this level? The plan with the other levels that have been completed is to put them on their own branch so we can go back and reference them later so we can use the content for inspiration/reference. Should I just delete the other levels now then merge this level into that branch?

sahansk2:
@benthayer Yeah, merging this level into that other branch sounds like a good idea.

PR #95, opened 2019-11-07
Ordering commits through timestamps

### LINKED ISSUE #25: When setting up level, order commits using timestamps
When we create the commits during the level setup process right now, they're all made at the same time and some of them end up with the same timestamp.  If we run git log, then commits show up out of order.

To fix, we can just set the time. When looking at the spec file, that should be the order of the timestamps and all timestamps should come before the time at which `create_tree` is called so that any commits created later on by the user don't end up with a timestamp before the commits Git Gud created

Relevant method is `create_tree` in gitgud/operations.py

sahansk2:
Aims to fix #25.

My first attempt at a fix involved modifying the arguments for self.repo.index.commit, but there's no visible timestamp change when `git log` is run. More still needs to be investigated at the time of this writing.


sahansk2:
Seems I spoke too soon -- it now seems to be working and putting increasing timestamps up to the current time for the commits. It just uses the for loop that the method is already using to iterate through the commits and assumes that the commits appear in the proper order through the loop.

bthayer2365:
Great, that is how you should be doing it. The file is guaranteed to be topologically sorted, which just means things are in their logical order. I just tested it, and it works great. 

Is there any way you can find the current timezone also? Currently they're all showing up with the +0000 timezone, but it should be whatever timezone the user is in so it looks like the time we expect.

sahansk2:
Sure, I can try to specify the timezone -- that might take a little more time, though, since the ISO 8601 format (what I'm feeding into gitpython right now) doesn't contain UTC information, but the RTC 2822 format (the other format that gitpython can make use of) does. 

I'll just have to figure out how to make the datetime object behave accordingly, and I'll commit once that's sorted.

bthayer2365:
The datetime.datetime object itself seems to have a tzinfo attribute. Maybe you can use the same code and just set the timezone 

https://docs.python.org/2/library/datetime.html#datetime.datetime.tzinfo

sahansk2:
Looking into GitPython, even though the ISO 8601 format does support UTC, GitPython will only accept formats that don't include it.
from [git/objects/util.py](https://github.com/gitpython-developers/GitPython/blob/master/git/objects/util.py):
```
def parse_date(string_date):
    """
    Parse the given date as one of the following
        * Git internal format: timestamp offset
        * RFC 2822: Thu, 07 Apr 2005 22:13:13 +0200.
        * ISO 8601 2005-04-07T22:13:13
            The T can be a space as well
```
The code that I've committed thus far just converts the datetime object into a string in iso format, so even if I specify the timezone through tzinfo, it's going to have to get dropped off in order to accommodate GitPython's parse_date() method.

This shouldn't be impossible to resolve -- all that's needed is just a bit of elbow grease to give GitPython the RFC 2822 it wants.

bthayer2365:
In the future, you may also want to try and refrain from using master branch as much as possible so you always have a clean branch to base other features off of.

bthayer2365:
Great, thanks for getting this done!

PR #105, opened 2019-11-08
Level/Skill Loading by Numbers

### LINKED ISSUE #26: Allow loading by level/challenge numbers

sahansk2:
Fixes #26.

So far, all that's been done is having the output formatting for `git gud skills` automated to a degree so that we don't have to directly modify the formatting to fix any changes if we add new skills/levels.

On the topic of level loading: I'm thinking that for loading by numbers, the user should be able to either run something like `git gud load 2` to load the second skill, or do `git git load 1-3` to load the third level for the 1st skill. 

What are your thoughts, @bthayer2365 ?


bthayer2365:
I think it makes sense to accept the following syntaxes:

```
git gud load 2
git gud load 1-3
git gud load 1 3
git gud load rampup
git gud load basics committing
```

Do you think we need anything else?

bthayer2365:
Usage would be as you described

bthayer2365:
Update, I also think we should allow mixing, like this:
```
git gud load basics committing
git gud load 1 1
git gud load basics 1
git gud load 1 committing
```

All four of these would have the same effect

sahansk2:
By how I've implemented this, on top of all those ways of loading that you mentioned, mixing with the hyphen works too!

For example, both `git gud load rampup-2` and `git gud load 2-relrefs1` load up `rampup relrefs1`.

bthayer2365:
Great. Tested it out, and it does everything it's supposed to - and more 

I'm going to add some cod that allows you to write some test cases for this, and I'll get back to you soon.

bthayer2365:
Alight, pull from master and you should be able to see the changes that I made. There's a spot for you to add the test in `gitgud/test_commands.py`. You can fill in test_load with whatever you want

sahansk2:
Through pytest, I iterated through all the possible valid combinations the user might input to load a level, and they all work!

I've modified `git gud skills` so that they show the skill numbers as well.

I also added a line that outputs on `git gud skills` which shows the user how to load a level. I *might* have overdone it though, so please tell me if I should make any changes.

bthayer2365:
First off, love the fancy text. It is a great addition! Can you open up a new issue and make a note to do this in more places, specifically colors. I'm not sure where else would be interesting, but my first instinct is to use it to differentiate the level names and stuff. If you'd like to handle the issue, say that you're working on it.

Alright, so now for the review. I actually saw this as soon as you commented, read the code, and started testing. I ended up running into some issues and ended up going down a rabbit hole of other issues, most of which were not yours, so thanks for the prompting. 

Long story short, I have a few issues with your code, but they're not that bad

The first issue is that your fancy text does not output correctly on Windows. It works on Linux, and I assume it works on Mac (not sure of your dev environment). It looks like this:

```
On CMD:
Load a level with [1mgit gud load[0m {[4mskill-number[0m|[4mskill-name[0m} [[[1m-[0m]{[4mlevel-number[0m|[4mlevel-name[0m}]

On Powershell:
Load a level with ?[1mgit gud load?[0m {?[4mskill-number?[0m|?[4mskill-name?[0m} [[?[1m-?[0m]{?[4mlevel-number?[0m|?[4mlevel-name?[0m}]
```

You can deal with this how you see fit, whether it is adding a check for windows and disabling fancy stuff or figuring out how to do fancy text on windows. The code might also benefit from using a library that can make fancy text easier. See if you can find one.

Also, I think it's best to put that fancy output in the help page for git gud load. You should keep the "load levels with git gud load" part though, since it's already a lot of text, and that additional text is useful. You can change the help message by editing the description parameter when the subparser is created. I also think you should be conscious of the other output that comes up when running `git gud help load`, but otherwise, do whatever you want.

The other issue I have is that I think your description doesn't accurately describe the usage. I think something like this would be better:
```
There are two usages:
git gud load {...}-[{...}]
git gud load {...} [{...}]

Ex:
git gud load basics 1
```

To address the progress I asked you to make, I think you should reduce the spacing so that the colons aren't aligned and instead do 4 spaces as the indent. I think the current separation is overemphasized and would look better the other way. Also, you forgot to push your test cases and I can't see them haha. 

Otherwise, numbers/other output looks good.

So:
0. Make new issue to do more fun escape sequences
1. Fix windows issue
2. See if there's a library to do fun stuff
3. Put output in help message for git gud 
4. List two usages
5. Upload test cases

sahansk2:
Oh! Sorry about the test cases, I deliberately excluded them from the commits. I'll commit them, but heads up, they're not exactly *elegantly* implemented -- but they get the job done.

I'll look into libraries that make color formatting easier, and hopefully they'll also have a mechanism to detect the OS running git-gud.

I'll go ahead and implement some of the other changes that you requested through commits one at a time.

sahansk2:
Something minor, though: I had to alter conftest.py because pytest-3 complained that "tmp_path" wasn't a valid fixture, so I replaced it with "tmpdir" which seemed similar enough, and it works. Should I commit that as well?

gitgud/__main__.py
```diff
@@ -222,14 +222,13 @@ def handle_skills(self, args):
...
             # TODO Add description
-            # 10 characters for the short IDs. 
             print(skill_format_template.format(i + 1, "\"" + skill.name + "\"", len(skill), ("", "s")[len(skill) > 1]))
-            
+
             for index, level in enumerate(skill):
-                # " " * (characters allocated for ID - 6)
                 print(level_format_template.format(index + 1, level.name))
         
         print("\nLoad a level with \033[1mgit gud load\033[0m {\033[4mskill-number\033[0m|\033[4mskill-name\033[0m} [[\033[1m-\033[0m]{\033[4mlevel-number\033[0m|\033[4mlevel-name\033[0m}]")
```
bthayer2365: Still doesn't work with windows

gitgud/test_commands.py
```diff
@@ -5,7 +5,31 @@
...
+
+                
+                subprocess.call(temp.format(skill.name, j + 1), shell=True)
+                assert all_skills[i][j] == gg.file_operator.get_level()
+        
+        subprocess.call('git gud load {}'.format(i + 1), shell=True)
+        assert all_skills[i][0] == gg.file_operator.get_level()
+        
+        subprocess.call('git gud load {}'.format(skill.name), shell=True)
+        assert all_skills[skill.name][0] == gg.file_operator.get_level()
```
bthayer2365: `test_load` takes a VERY long time. You are loading/unloading every level four times. That is overkill. You just need to test each thing once, so you only need to do each of these situations only once, no iteration at all:
```
git gud load number
git gud load name
git gud load number name
git gud load name number
git gud load number-name
git gud load name-number
```
This should be the simplest thing we can test. It does one of each "half-situation", with number/name on each side of the space/hyphen.

bthayer2365:
> 
> 
> Something minor, though: I had to alter conftest.py because pytest-3 complained that "tmp_path" wasn't a valid fixture, so I replaced it with "tmpdir" which seemed similar enough, and it works. Should I commit that as well?

Interesting. Are you running on mac? I'm not sure why that wouldn't be a valid fixture. It works on both of my windows computers and on the docker image that runs the tests

bthayer2365:
Also, if you'd like, you can merge master and see if that code works as expected. I've fully updated the test suites so all test cases pass on the CI server, maybe this will fix the issue for you

bthayer2365:
At this point, it might be safest to do a hard reset to bcbe, and start the merge again from scratch

bthayer2365:
We can check out 1e90 if for some reason we need that commit again

sahansk2:
Well, seems like the merge got sorted out -- I think the issue was that I tried to resolve the merge both in the website, and through VSCode. Sorry if you got bombarded with email notifications -- I still have to get used to resolving merge conflicts, it seems. 

Also, regarding the tmp_path/tmpdir thing: I'm running on Ubuntu right now, and so I'm actually running "pytest-3", since "pytest" points to python2 and not python3. Here's the output of `pytest-3 --version`:

`This is pytest version 3.3.2, imported from /usr/lib/python3/dist-packages/pytest.py`

I still have to use tmpdir, but it's not that big a deal for me.

bthayer2365:
Can you look into why that fixture isn't working? If not, I can. The point of tests is to be able to run things more smoothly, so we should resolve issues like that.

Don't worry, all the notifications come on the same email chain, and I normally have my phone silent, so it's no trouble

sahansk2:
I found out that the reason the 'tmp_path' fixture isn't working is that it was [introduced in version 3.9 of pytest. ](http://doc.pytest.org/en/latest/changelog.html#pytest-3-9-1-2018-10-16). The version in the Ubuntu repositories was 3.3, so I didn't have tmp_path as a valid fixture.

bthayer2365:
Okay. Can you change and push, we'll see if the build succeeds

sahansk2:
I think the necessary changes have been made now -- I've put one load for every possible type of string, and I've stashed away the color formatting thing for a different PR.
Take a look, @benthayer , and tell me if anything more needs to be done.

gitgud/__main__.py
```diff
@@ -253,20 +258,31 @@ def handle_levels(self, args):
...
         self.assert_initialized(skip_level_check=True)
-        if args.skill_name in all_skills:
-            skill = all_skills[args.skill_name]
+        
+        argset = args.skill_name.split("-")
 
+        if (len(argset) == 2):
+            args.skill_name, args.level_name = tuple(argset)
+        elif(len(argset) > 2):
+            print("You put too many dashes! Use at most one dash when trying to load a level.")
```
benthayer: Too many dashes is a weird error, do you think you could find the index of the dash and then do
```
dash_index = whatever
args.skill_name = args.skill_name[:dash_index]
args.level_name = args.skill_name[dash_index]
``` 
should then throw an error like
```
$ git gud load basic-s-committing
No skill "basic"
```
```
$ git gud load basics-s-committing
No level "s-committing"
```

benthayer: The error handling is already implemented, you'd just be passing the bad information along

gitgud/__main__.py
```diff
@@ -253,20 +258,31 @@ def handle_levels(self, args):
...
+            return
+        
+        args.skill_name = int(args.skill_name) - 1 if args.skill_name.isnumeric() else args.skill_name
+        args.level_name = int(args.level_name) - 1 if (args.level_name is not None and args.level_name.isnumeric()) else args.level_name
```
benthayer: the `X if cond else Y` syntax is supposed to make things more easy to understand, do you think could change this? See comment in general review for more suggestions

gitgud/test_commands.py
```diff
@@ -5,7 +5,17 @@
...
-    assert level == gg.file_operator.get_level()
\ No newline at end of file
+    load_tests = [
+        ('git gud load 1', all_skills[0][0]),
+        ('git gud load rampup', all_skills["rampup"][0]),
+        ('git gud load 2 detaching', all_skills[1]["detaching"]),
+        ('git gud load rampup 4', all_skills["rampup"][3]),
+        ('git gud load 3-octopus', all_skills[2]["octopus"]),
+        ('git gud load basics-4', all_skills["basics"][3])
+    ]
```
benthayer: Lest add the single test case of  `git gud load -2`

benthayer:
Can you also make sure to update the description of load_parser and the readme to match the new usage

sahansk2:
This works, but I wasn't able to implement quite what you were saying.

If we're going to be using the `keys()` method instead of `__contains__()`, then we'll have to either hardcode what's going to be returned or set it up as a function taking a parameter. There are two different "sets" of keys, I guess you could say: one has all the names (`'basics', 'rampup', 'extras'`, etc.), and one has the indexes that map the names to the actual items (`0, 1, 2, 3`, ...). 

The implementation that I used here assumes that `keys()` should always return that second set with the indexes, but depending on what type is passed into it, it'll either return the incremented indexes (for what the user inputs directly) or the indexes as they are (for internal usage).

I thought it might be a little strange to return either the names or the indexes as a function of whatever's being checked (e.g. running `if 'rampup' in all_skills.keys('rampup')`, and then having `keys()` return the list of the names because `'rampup'` isn't numeric), so for now, I've put everything in `__contains__()` since using the `in` operator directly seems a bit less obscure.

Should I go ahead and set it up so that `keys()` takes another parameter, and move all the related functionality from `__contains__()` and into `keys()`?

gitgud/skills/util.py
```diff
@@ -204,6 +204,10 @@ def __getitem__(self, query):
         if isinstance(query, int):
             return self._items[query]
         elif isinstance(query, str):
+            if query.isnumeric():
+                if int(query) == 0:
```
benthayer: Can you make this more broad,  0 < int(query) <= len(self)

gitgud/skills/util.py
```diff
@@ -223,6 +227,8 @@ def __setitem__(self, key, item):
 
     def __contains__(self, key):
```
benthayer: The only line we need here is `key in self._items`

gitgud/skills/util.py
```diff
@@ -204,6 +204,10 @@ def __getitem__(self, query):
         if isinstance(query, int):
             return self._items[query]
```
benthayer: We don't need to accept ints anymore

README.md
```diff
@@ -90,7 +90,7 @@ Some commands you'll want to learn and what they do:
   * Displays all levels in the specified skill
 * git gud load <skill_name>
   * Begin the first level of the specified skill
-* git gud load <skill_name> <level_name>
+* git gud load <skill_name> <level_name> / git gud load <skill_name>-<level_name> / git gud load <skill_name> / git gud load -<level_name>
   * Begin the specified level (it is optional to specify the level)
```
benthayer: What about just saying (in fewer words) on this indented line that a hyphen can be used in place of the space and if a hyphen is used and the skill is omitted, then it defaults to the current skill

gitgud/__main__.py
```diff
@@ -45,7 +45,7 @@ def __init__(self):
         progress_parser = self.subparsers.add_parser('progress', help='Continue to the next level', description='Continue to the next level')
         skills_parser = self.subparsers.add_parser('skills', help='List skills', description='List skills')
         levels_parser = self.subparsers.add_parser('levels', help='List levels in a skill', description='List the levels in the specified skill or in the current skill if Git Gud has been initialized and no skill is provided.')
-        load_parser = self.subparsers.add_parser('load', help='Load a specific skill or level', description='Load a specific skill or level')
```
benthayer: The description here is meant to be really long. Feel free to take up multiple lines and do a full explanation of how to use this command

gitgud/skills/util.py
```diff
@@ -201,13 +201,12 @@ def __init__(self, names, items):
...
             if query.isnumeric():
-                if int(query) == 0:
+                if 0 < int(query) <= len(self):
+                    return self._items[int(query) - 1]
+                else:
                     raise KeyError
-                return self._items[int(query) - 1]
             return self._items[self._name_dict[query]]
         else:
             raise ValueError('Bad key type.')
```
benthayer: Since we're only accepting strings now, it makes sense to call this a KeyError with no arguments just like we do on above.

gitgud/skills/util.py
```diff
@@ -225,30 +224,20 @@ def __setitem__(self, key, item):
...
 
     def values(self):
         return self._items
     
-    def keys(self, keytype): # return the numbers, or return the names?
-        if keytype is str:
-            return {keytype(int(k) + 1) for k in self._name_dict.values()}
-        return {keytype(k) for k in self._name_dict.values()}
+    def keys(self): 
+        return set([str(int(k) + 1) for k in self._name_dict.values()]) | set(self._name_dict.keys())
```
benthayer: I think it would make sense to separate this out into three lines. The three lines would be setting a variable for the first set and a variable for the second set, then finally combining them in the return statement. The variable names should be descriptive

 Also instead of  using `self._name_dict.values`, why not use `range(len(self))` then do the i + 1 stuff. Also, use i instead of k since it's an iteration parameter, not a key.

gitgud/test_commands.py
```diff
@@ -6,13 +6,13 @@
...
-        ('git gud load 3-octopus', all_skills[2]["octopus"]),
-        ('git gud load basics-4', all_skills["basics"][3]),
-        ('git gud load -2', all_skills["basics"][1])
+        ('git gud load 1', all_skills["1"]["1"]),
+        ('git gud load rampup', all_skills["rampup"]["1"]),
+        ('git gud load 2 detaching', all_skills["2"]["detaching"]),
+        ('git gud load rampup 4', all_skills["rampup"]["4"]),
+        ('git gud load 3-octopus', all_skills["3"]["octopus"]),
+        ('git gud load basics-4', all_skills["basics"]["4"]),
+        ('git gud load -2', all_skills["basics"]["2"])
```
benthayer: I changed this test case to refer to "rampup" just to make sure that it's not defaulting to the first level

benthayer:
You'll also want to make sure to format the code so it's pretty

https://code.visualstudio.com/docs/python/linting

benthayer:
See if you can resolve those merge conflicts too

README.md
```diff
@@ -90,8 +90,11 @@ Some commands you'll want to learn and what they do:
   * Displays all levels in the specified skill
 * git gud load <skill_name>
   * Begin the first level of the specified skill
+* git gud load -<level_name>
+  * Begin the specified level of the current skill
 * git gud load <skill_name> <level_name>
-  * Begin the specified level (it is optional to specify the level)
+  * Begin the specified level (specifying the level is optional)
```
benthayer: Since we have the other git gud load <skill_name> above, we don't need to say it's optional

README.md
```diff
@@ -90,8 +90,11 @@ Some commands you'll want to learn and what they do:
   * Displays all levels in the specified skill
 * git gud load <skill_name>
   * Begin the first level of the specified skill
+* git gud load -<level_name>
+  * Begin the specified level of the current skill
 * git gud load <skill_name> <level_name>
-  * Begin the specified level (it is optional to specify the level)
+  * Begin the specified level (specifying the level is optional)
+  * A hyphen can be used instead of a space between <skill_name> and <level_name> (specifying the skill is optional)
```
benthayer: Don't say that specifying the skill is optional. We already have the other thing above

sahansk2:
Yikes, I didn't commit the changes to test_util.py which takes into consideration that `int` keys can't be used in NamedList objects. All the tests should be passing now.

Just as an aside, is there a certain file that I can run pytest on to run all the tests in the whole project? Right now, I'm manually locating each test python script file and running them individually, but I'd like to know if that's how it's meant to be done right now.

benthayer:
You can run `pytest .` from the top-level directory to run all the tests.
I'll review this when I can

benthayer:
From now on, CircleCi will automatically run all the test for you on GitHub and we can see the results here. This will also be true of any pull requests you open in the future. As you can see, tests are passing!

gitgud/__main__.py
```diff
@@ -30,14 +30,38 @@ def __init__(self):
...
+            "   git gud load -3",
+            "\n",
+            "======Loading the first level on a skill\n",
+            "This command loads the first level on the specified skill:\n",
+            "   git gud load <skill>\n",
+            "<skill> may be either the name or the number of the skill.",
+            "Here are some example uses, which both load the first level of the first skill:\n",
+            "   git gud load basics",
+            "   git gud load 1",
+        ])
```
benthayer: This might be too long, especially when looking at it all on the command line. Do you think you could get it down to 15-20 lines?

You seem to be restating things towards the end. It's good to be clear, but at that point we're sacrificing usability for clarity. This is so long that users would have to scroll for a bit to get back to where they were. People should be able to figure it out with just a few examples. 

The shorter the better, just make sure it's understandable

sahansk2:
I fixed the README issues from earlier with mentioning stuff was optional, and I trimmed down the load_description.

I removed the examples for the `Additional Commands` section since I assumed that if a user is comfortable with how to use the normal commands, figuring out the additional commands should be no problem. Furthermore, I rewrote the description string so that the appearance in-file is similar to how it would appear in the console -- tell me if anything more needs to be changed.

benthayer:
Do you think you can go shorter?

(I added more equals signs because I thought it looked nicer)

sahansk2:
It's been discussed before, but I'll mention it here just so that there's a record of it...
It's pretty difficult to crop this down even further without losing important info. The only way that I could reduce it is by cutting down the number of newlines, but that sacrifices readability.

benthayer:
Okay, will look at this on Monday. Should be good to merge. Will bring this up again after user testing

PR #111, opened 2019-11-12
Tests for NamedList class

### LINKED ISSUE #106: Extend test_access in `test_util.py` to include all methods defined in NamedList
Each dunder method should be covered.

sahansk2:
Fixes #106

### DIFF
--- a/gitgud/skills/test_util.py
+++ b/gitgud/skills/test_util.py
@@ -29,25 +29,31 @@ def test_goal():
         for level in skill:
             os.path.isfile(level.goal_path)
 
-def test_contains_NL():
-    assert all_skills[1] in all_skills
-    assert all_skills[1][2] in all_skills[1]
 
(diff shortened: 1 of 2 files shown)


gitgud/skills/test_util.py
```diff
@@ -28,3 +28,32 @@ def test_goal():
...
+    assert nltest[0] == 5
+    assert nltest['baz'] == 9
+
+
+def test_iter_NL():
+    ['' for skill in all_skills]
+
+
+def test_len_NL():
+    assert len(all_skills[0]) == 4
```
bthayer2365: If we add a level, we have to change this, but that might get annoying. Can you change this to use `named_list`

bthayer2365: If we add a level, we have to change this, but that might get annoying. Can you change this to use `nltest` like you do in the other methods

PR #113, opened 2019-11-13
Updated the welcome message to include info on some basic commands & …

### LINKED ISSUE #109: Add a welcome message when git gud is first initialized
Right now, it just says "Welcome to Git Gud.", but I would love if we had more information listed that directs users to this repo and explains the Git Gud commands commonly used.

attard-andrew:
fixes #109

…a link to the Git repo.

@bthayer2365 - Updated the welcome message.


attard-andrew:
@benthayer Sure - I'll give that a go.

attard-andrew:
@benthayer I think that should do it.

gitgud/__main__.py
```diff
@@ -159,27 +159,7 @@ def handle_init(self, args):
...
-                 git gud test: Once you've completed the current level's goals,
-                 use this to check completion.
-
-                 git gud progress: Whether you've completed the current level or
-                 just want to move on to the next level, use this to proceed!
-
-                 For more information on Git Gud, including a full list of 
-                 commands, checkout https://github.com/bthayer2365/git-gud
-        ''')
+        print(open('welcome.txt').read())
```
benthayer: See `skills/basics/__init__.py` for how to add files. Python looks for this document where we run the command, not where the package is installed.

attard-andrew: Alright well I ran pytest.py - and I can see the file not found errors. Looking at __init__.py, I assume - somewhere. I would need to add a line similar to:

BasicLevel('welcome', pkg_resources.resource_filename(__name__, '_welcome/'))

However - I honestly have no understanding of what 'packages' are in this context or how this works, so I'll have to do some Googling to figure things out first

benthayer: Not sure if you've figured it out yet, but packages are a collection of modules. Modules are basically just python files. So, `gitgud` is a package as is `gitgud.skills`. The package we'll be referring to is `gitgud`, which is what `__name__` will refer to if you use it on this line. 

You can use print statements to see what's going on

attard-andrew: Hey @benthayer - sorry for the slow follow up, been doing some traveling :).

So I understand the basic idea behind packages, but what I don't understand is how or why the new text file needs to be added to the package - is it not included as part of the package by being in the directory & being referenced in .py? If I try to piece it together..

- It kind of looks like `__init__.py` in the 'skills' directory just imports all of the modules within that directory's subfolders (basics, rampup, extras).  When importing the name of those referenced folders are updated using "as basics_skill" - why is this renaming necessary before assigning them in the array below to be imported?? or is it?
- If I then look at the `__init___.py` in the 'basics' subdirectory, it looks like this is again just importing all of the subdirectories containing the different 'basics' skills
- I noticed that 'util' is mentioned here, so I opened that to take a look at "BasicLevel" there.  I haven't learned anything about classes yet - but it kind of looks like you're using it to define a standard structure for each basic level and it's searching for files named in a standardized format to define the instructions/goal text etc. - Does this mean you've got another class somewhere else
- Does this mean that the import definition can refer to a directory - as I THINK you've with:

`BasicLevel('committing', pkg_resources.resource_filename(__name__, '_committing/')),`

OR a specific file (such as my welcome.txt)???

- The `__init__.py` in the 'gitgud' directory so far makes no sense to me

benthayer: Hey @attard-andrew, I hope you've been having fun!

So a bit of clarification, text files are never "added" to packages, they can only be accessed. 

Imports will never refer to a file, unless it's a `.py` file, in which case it's known as a "module". This basically means that we have to look up the exact file path using something like `pkg_resources.resource_filename(__name__, '_committing/')`, or in our case, `pkg_resources.resource_filename(__name__, 'welcome.txt')`. We can't do `open("welcome.txt")` because python will think that it's referring to the directory that the user is running in. `pkg_resources` is just the name of a package that we use to help us do this (since we are a package, and it helps us with things related to resources).

On to `__init__.py`...
We rename the skills because otherwise, they would all have the same name. `as X` just defines something to have the name `X` for when we use it. The only renaming is occurring in the `__init__.py` file. The purpose of any `__init__.py` file is to allow names to be imported from that module. `from gitgud.skills.extras import skill as extras_skill` wouldn't be possible if we didn't define `skill` in `gitgud/skills/extras/__init__.py`. Python will automatically know how to find modules and packages, but the things we define in `__init__.py` can be anything including variables, classes or whatever else. 

Now, when we're running `BasicLevel('committing', pkg_resources.resource_filename(__name__, '_committing/')),`, there's no importing being done there. We're defining the name of the level, "committing", and the path to the files. As I said, `pkg_resources.resource_filename(__name__, '_committing/')` looks up the location of `__name__`, which refers to the current module, on the filesystem, and adds on `_committing/` so we have the right path.

Let me know if all this makes sense

benthayer:
@attard-andrew How's it going? If you're still confused and would like to find a time to chat, we can do that

PR #123, opened 2019-12-19
Track commits in .git/gud/commits.json

### LINKED ISSUE #40: Keep track of commit hashes using a .json file
We should keep track of the hashes of the commits we create using `git gud commit` and `git gud load`. Each one of these commits will a name, as defined in #34, but we should create a file, `.git/gud/commits.json` and store the hashes of the commits that we create. This will be useful for checking if the level is complete and validating what we think happened happened. The json should look like this:
```
{
'commit_name': 'hash'
...
}
```
Where the hash is what shows up when we run `git log` and `commit_name` is something like `C1`, `C2`, `M1`

Don't forget about Python's `json` module

parkmichael777:
fixes #40


parkmichael777:
#40 

Here's what happens as of now:
1. When you load a skill, load a level, or reset a level, the .git/gud/commits.json file is overwritten with a fresh file. This file is populated with whatever commits the level is initialized with.
2. Whenever a user uses "git gud commit" the commits.json file will add a dictionary entry.

commit_name => "C#"
commit_hash => C#'s commit hash.

I haven't built any functions to read the JSON file, but that should be easy enough.

gitgud/operations.py
```diff
@@ -95,7 +97,9 @@ def create_tree(self, commits, head):
                 for parent in parents[1:]:
                     merge_base = self.repo.merge_base(parents[0], parent)
                     self.repo.index.merge_tree(parent, base=merge_base)
-                self.repo.index.commit(name, author=actor, committer=actor, author_date=committime_rfc, commit_date=committime_rfc, parent_commits=parents)
+                commit_obj = self.repo.index.commit(name, author=actor, committer=actor, author_date=committime_rfc, commit_date=committime_rfc, parent_commits=parents)
+                track_commits(name, commit_obj.hexsha[:7])
+                
 
             commit_objects[name] = self.repo.head.commit
```
benthayer: Do you want to change this to use your new `commit_obj`? It refers to the same thing.

gitgud/operations.py
```diff
@@ -215,3 +219,22 @@ def get_operator():
...
+def track_commits(commit_num, commit_hash):
+    # Assumes that .git/gud/commits.json has been initialized by 'git gud load'
+    if os.path.exists(".git/gud/commits.json"):
+        with open(".git/gud/commits.json") as fp:
+            commit_dict = json.load(fp)
+        commit_dict["C" + commit_num] = commit_hash
+        with open(".git/gud/commits.json", 'w') as fp:
+            json.dump(commit_dict, fp)
+    else:
+        print("uh oh someting borken")
```
benthayer: For all of these, accept a file operator as a parameter and then use it to get the file paths. If we create a directory and switch to into it (which might be the case with later levels), this won't work

gitgud/operations.py
```diff
@@ -215,3 +219,22 @@ def get_operator():
...
+
+def init_tracking_json():
+    if os.path.exists(os.path.join(os.getcwd(), ".git")):
+        with open(".git/gud/commits.json", 'w') as fp:
+            json.dump({}, fp)
+    else:
+        print("ERROR: Wrong directory :C")
+            
+
+def track_commits(commit_num, commit_hash):
```
benthayer: call this `track_commit` singular

benthayer:
We don't need the reading just yet. It's not clear what we're going to need this for, so we can hold off on  doing that until we start running into issues and need this.

We'll probably need this when working with rebases and merges and trying to figure out if the user has solved a given level

gitgud/operations.py
parkmichael777: check

gitgud/operations.py
parkmichael777: done

gitgud/operations.py
parkmichael777: bingo

parkmichael777:
I need to reset the tracking json whenever there is a new level; I do this for load, reset, and init. Are there any more that I don't know about or am forgetting?

gitgud/operations.py
```diff
@@ -221,12 +221,11 @@ def get_operator():
...
 
 def init_tracking_json(file_operator):
-    if os.path.exists(os.path.join(file_operator.git_path, "gud", "commits.json")):
-        with open(os.path.join(file_operator.git_path, "gud", "commits.json"), 'w') as fp:
-            json.dump({}, fp)
-    else:
-        print("ERROR: Wrong directory :C")
-   
+    print("Creating: Commit Tracker")
+    with open(os.path.join(file_operator.git_path, "gud", "commits.json"), 'w') as fp:
```
benthayer: You can change this to file_operator.git_gud_path, but I would add an attribute to file_operator so you can do file_operator.commit_data_path

parkmichael777: Fixed in #130

gitgud/__main__.py
```diff
@@ -160,6 +162,7 @@ def handle_init(self, args):
         print('Welcome to Git Gud!')
         print()
 
+        init_tracking_json(self.file_operator)
         self.file_operator.get_level().setup(self.file_operator)
         show_tree()
```
benthayer: Remove these three lines and use `load_level` function here. `load_level` will needlessly write the level again, but that's a minor cost, and the change would improve readability.

benthayer: Change these three lines to use `load_level`. It'll needlessly rewrite the level to the file, but that doesn't change anything, the new code makes the code more maintainable.

`init_tracking_json` will be in the `load_level` function

gitgud/__main__.py
```diff
@@ -188,6 +191,7 @@ def handle_reset(self, args):
 
         level = self.file_operator.get_level()
         print("Resetting...")
+        init_tracking_json(self.file_operator)
         level.setup(self.file_operator)
         show_tree()
```
benthayer: Use `load_level` here

gitgud/__main__.py
```diff
@@ -264,12 +267,14 @@ def handle_load(self, args):
 
             if args.level_name is not None:
                 if args.level_name in all_skills[args.skill_name]:
+                    init_tracking_json(self.file_operator)
```
benthayer: You'll be able to remove this once `init_tracking_json` is in `load_level`

gitgud/__main__.py
```diff
@@ -264,12 +267,14 @@ def handle_load(self, args):
...
                     level = skill[args.level_name]
                     self.load_level(level)
                 else:
                     print('Level "{}" does not exist'.format(args.level_name))
                     print("To view levels/skills, use git gud levels or git gud skills")
             else:
+                init_tracking_json(self.file_operator)
```
benthayer: Same here, you can remove this once `init_tracking_json is in `load_level`

gitgud/__main__.py
```diff
@@ -293,6 +298,7 @@ def handle_commit(self, args):
         print('Simulating: git commit -m "{}"'.format(commit_name))
 
         commit = self.file_operator.add_and_commit(commit_name)
+        track_commit(self.file_operator, commit_name, commit.hexsha[:7])
```
benthayer: This is currently the only place where we use `add_and_commit` currently, and I think in the future, we'll want to track every commit that we create with the `add_and_commit` function, so put `track_commit` in `add_and_commit`

gitgud/operations.py
```diff
@@ -215,3 +219,23 @@ def get_operator():
...
             return Operator(path)
     return None
+
+
+def init_tracking_json(file_operator):
+    print("Creating: Commit Tracker")
+    with open(os.path.join(file_operator.git_path, "gud", "commits.json"), 'w') as fp:
+        json.dump({}, fp)
+        print("Successfully created")
+        print()
```
benthayer: Remove all logging, this is an internal function, so seeing this isn't useful to the user.

gitgud/__main__.py
parkmichael777: added init_tracking_json to load_level. Deleting these three lines.

parkmichael777: Same as above, yes?

gitgud/__main__.py
parkmichael777: Replaced those three lines with load_level.

gitgud/__main__.py
parkmichael777: If i'm not mistaken, you got rid of this whole section in the commit that you made me merge, right?

gitgud/__main__.py
parkmichael777: Also a part of that section you deleted.

gitgud/__main__.py
parkmichael777: Removed track_commit from handle_commit, moved to add_and_commit.

gitgud/operations.py
parkmichael777: Logging removed

parkmichael777:
All requested changes have been made. 

In regards to the "test," are you talking about the CircleCI tests? Or a git gud test for the user?

PR #128, opened 2019-12-31
Fix hook permissions, progress on #79

sahansk2:
Fixes #125 and fixes #77, which are duplicates. Credit goes to @divCurlF for this PR.

This issue has been brought up a second time, and since @divCurlF did most of the fixes already, I just went ahead and implemented them (meaning the rest of them).

Truth to be told, I'm not as confident resolving these merge conflicts, since I didn't author the code, but I can investigate and try resolving them anyway.

PR #131, opened 2020-01-01
Abort commit

parkmichael777:
This code uses the pre-commit hook to prevent commits via 'git commit <name>'. It also prints a message pointing the user to git gud commit and informing them that their commit has been aborted.

This code is dependent on hooks being skipped by gitpython commits (so that git gud commit will still commit)

Furthermore, for future use of the pre-commit hook (although, I don't see how it can be used considering the changes in this PR and #130), the bash script needs to be modified so that an 'exit 0' condition can be reached.


parkmichael777:
The only new commits for this feature is the last three (b8727f5, cec9e9a, and c0769a9)

gitgud/hooks/precommit.py
```diff
@@ -0,0 +1 @@
+print("Error: Please use 'git gud commit'. Aborting your commit now.")
```
benthayer: Can you change this to 
```
print("Git Gud: git commit is disabled for this level. Please use 'git gud commit'.")
```

gitgud/hooks/__init__.py
```diff
@@ -3,5 +3,6 @@
     # ('git-hook-name', 'modulename')
     ('post-rewrite', 'postrewrite'),
     ('commit-msg', 'commitmsg'),
-    ('post-merge', 'postmerge')
+    ('post-merge', 'postmerge'),
```
benthayer: Go back through this branch and do an interactive rebase to get rid of all the code added by post-merge. You can force push to here

gitgud/__main__.py
```diff
@@ -155,7 +155,10 @@ def handle_init(self, args):
                 hook_file.write('#!/bin/sh' + os.linesep)
                 hook_file.write('cat - | '
                                 + python_exec + ' -m gitgud.hooks.' + module_hook_name + ' "$@"' + os.linesep)
-                hook_file.write('exit 0' + os.linesep)
+                if git_hook_name == "pre-commit":
+                    hook_file.write('exit 1' + os.linesep)
+                else:
+                    hook_file.write('exit 0' + os.linesep)
```
benthayer: This is not right.

The idea is that we can use our python hook (`precommit.py`) to give the exit code, then we don't need to clutter `handle_init`. Just add `exit(1)` to `precommit.py`

gitgud/hooks/precommit.py
parkmichael777: Changed.

gitgud/hooks/__init__.py
parkmichael777: All features are rebased off track_commits. Track_commits has been modified to hold the definitive versions of shared files.

gitgud/__main__.py
parkmichael777: Ah, I understand now. This was written before I knew about #128. I will add the correct exit code but be sure to merge sahansk2's pull request—It will not work until that code has been integrated.

gitgud/hooks/precommit.py
```diff
@@ -1,5 +1,4 @@
 import sys
 
-print('Command "git commit" is diabled. Use "git gud commit" instead.')
+print('Command "git commit" is disabled. Use "git gud commit" instead.')
 sys.exit(1)
```
benthayer: You can just remove the import statement and change `sys.exit(1)` to`exit(1)`. `exit` is a built-in

gitgud/operations.py
```diff
@@ -34,6 +36,7 @@ def add_and_commit(self, name):
         # TODO Commits with the same time have arbitrary order when using git log, set time of commit to fix
         self.add_file_to_index(name)
         commit = self.repo.index.commit(name, author=actor, committer=actor, skip_hooks=True)
+        track_commit(self, name, commit.hexsha[:7])
```
benthayer: Why do we have this here, isn't this out of scope for this PR?

gitgud/operations.py
```diff
@@ -21,6 +22,7 @@ def __init__(self, path, initialize_repo=True):
             self.repo = None
 
         self.git_path = os.path.join(self.path, '.git')
+        self.commits_json_path = os.path.join(self.path, '.git', 'gud', 'commits.json')
```
benthayer: Isn't this out of scope for this PR?

gitgud/operations.py
```diff
@@ -215,3 +221,20 @@ def get_operator():
...
+def track_commit(file_operator, commit_num, commit_hash):
+    # Assumes that .git/gud/commits.json has been initialized by 'git gud load'
+    if os.path.exists(os.path.join(file_operator.git_path, "gud", "commits.json")):
+        with open(os.path.join(file_operator.git_path, "gud", "commits.json")) as fp:
+            commit_dict = json.load(fp)
+        commit_dict["C" + commit_num] = commit_hash
+        with open(os.path.join(file_operator.git_path, "gud", "commits.json"), 'w') as fp:
+            json.dump(commit_dict, fp)
+    else:
+        print("ERROR: Commit tracker does not exist!")
```
benthayer: These seem like the belong in other PRs, is that correct?

gitgud/hooks/precommit.py
parkmichael777: done

gitgud/operations.py
parkmichael777: It's there because all features (abort_commit, track_merges, track_rebases) are based off the track_commits branch.

gitgud/operations.py
parkmichael777: It's there because all features (abort_commit, track_merges, track_rebases) are based off the track_commits branch. There are certain changes (like the hooks/init file, hooks paths, and __main__ formatting) that need to propagate to all feature branches. Whenever these need changing, they are done to the track_commits branch and then the other branches are rebased on top of track_commits. This comes with the side effect that track_commits stuff gets added, but most of that is pretty useful for debugging so I didn't mind it propagating.

gitgud/operations.py
parkmichael777: yes specifically track_commits.

parkmichael777:
Ah, I realize now since you are using github's website to view all the additions/deletions, it might not have been the best idea to do the rebases because it makes it seem like foundational code changes (changes made in track_commits) are specific to abort_commit. Sorry about that. It might be useful to look at the tree.

benthayer:
Since the changes are unneccessary for this PR and I'd rather review the changes in the track_commits branch, can you delete that code? We can add it back later. I don't mind it being a manual deletion as opposed to a revert/rebase, I'm probably going to squash this anyways.

parkmichael777:
@benthayer Ok, I deleted it. I think I got most of it.

PR #132, opened 2020-01-03
Name and Track rebases

### LINKED ISSUE #43: Complete the `post-rewrite` hook to keep track which commits have been rebased

parkmichael777:
fixes #43 
May need to clean up logging.

Will need to add ability to add user flags.


parkmichael777:
Branches

New commits start from 625c26a

PR #90, opened 2019-11-04
Implement `git gud load next`

TSSlade:
fixes #31

PR #134, opened 2020-01-10
Implements `git gud load next`

### LINKED ISSUE #133: `git gud load prev` & `git gud load previous`
Similar to #31, but we should be able to go backwards too. 

If you accidentally load the next level or just want to go back, this would be convenient

sahansk2:
Fixes #90 
fixes #133

### DIFF
--- a/gitgud/__main__.py
+++ b/gitgud/__main__.py
@@ -295,25 +295,25 @@ class GitGud:
 
         skill_to_load = self.file_operator.get_level().skill.name
         if args.skill_name:
-            if args.skill_name.lower() == "next":
-                print("Progressing to next skill...")
-                print()
+            if args.skill_name.lower() in {"next", "prev", "previous"}:
+                query = args.skill_name.lower()
                 level = self.file_operator.get_level()
-                next_level = level.next_level
-                if next_level is not None:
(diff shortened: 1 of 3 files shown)


sahansk2:
Now fixes #133 as well.
I originally thought the code for loading previous/prev was going to be different, with an O(n) search for the skill/level names, but it ended up being much easier.

There's repeated code, so I'm going to make this cleaner.

gitgud/__main__.py
```diff
@@ -304,14 +288,35 @@ def handle_load(self, args):
...
+                    level_to_load = level.prev_level
+                
+                if level_to_load is not None:
+                    self.load_level(level_to_load)
+                else:
+                    if query == "next":
+                        print_all_complete()
+                    else:
+                        print("Already on the first level; reloading first level...\n")
+                        self.load_level(all_skills["1"]["1"])
```
benthayer: Can you remove this and change the print statement above to 
```
print('Already on the first level. To reload the current level, use "git gud reload"')
```

sahansk2:
I've changed the print text, and how `git gud load prev/previous` works on the first level -- this PR should be good now.

benthayer:
Great, I'll check it

PR #142, opened 2020-01-25
Removed mentions of git gud progress

sahansk2:
This should have been part of #134 -- all mentions of the old command are gone now.


benthayer:
Great, thank you

Issue #145, opened 2020-02-19
Add CONTRIBUTORS file to repository


e-leclercq:
Add an exhaustive list of contributors to the **git-gud** repository in the form of an html file, `CONTRIBUTORS`. Later, include information about this file in the `AUTHORS` section of the **git-gud** man page #144.

benthayer:
How does this look in other repositories? I thought it was standard to link to GitHub, which will always show an up to date list of contributors

PR #146, opened 2020-02-19
Added manpage for git-gud

### LINKED ISSUE #144: Add man page for `git gud`
We should write a man page for `git-gud`. This feature would be a good way to tackle issue #109.

e-leclercq:
Fixed #144 
Wrote and added a manpage for git-gud. Will possibly submit another PR soon to add this to the setup.py script, so it is installed in the correct `man` directory.


git-gud.1
```diff
@@ -0,0 +1,68 @@
+.TH git-gud 1 "19 Feb 2020" "version 0.0"
+.SH NAME
+git-gud \- get good at git with git-gud
```
benthayer: Can we capitalize "Get good..."

git-gud.1
```diff
@@ -0,0 +1,68 @@
...
+.SH SYNOPSIS
+.B git gud
+[option1] [option2]
+.SH DESCRIPTION
+.B git-gud
+teaches you how to use Git with fun interactive tutorials.
```
benthayer: "Teaches..."

git-gud.1
```diff
@@ -0,0 +1,68 @@
...
+Use the following command to create a Git repository and start the first level:
+.B python\ \-m\ gitgud\ start
```
benthayer: We don't really use this (we can, but we don't advertise it). Git gud is actually registered as a program. If you look in wherever you have python installed, you'll see a `git-gud` executable which is run when you type in `git gud`

Also, the start command is out of date. It's `git gud init`.

git-gud.1
```diff
@@ -0,0 +1,68 @@
...
+.TP
+\fBgoal\fP
+Concisely prints out what is needed to complete the level.
+.TP
+\fBshow\-tree\fP
+Visualizes the current git tree.
+.SH BUGS
+If you believe you have found a bug in \fBgit-gud\fR, please file an issue at this address https://github.com/benthayer/git-gud/issues/
+.SH AUTHOR
+Written by Evan LeClercq (eleclercq@tuta.io).
```
benthayer: Can you add a link to the github page here so people can find all the authors of git gud themselves. Give me explicit credit for starting it, and mention OSAI. It's also fine to give yourself explicit credit for the man page as you've done.

It'd probably make sense to follow the style of the git man page for this section

git-gud.1
```diff
@@ -0,0 +1,68 @@
...
+.SH SEE ALSO
+.TP
+git(1)
+.SH DISCLAIMER
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+See the LICENSE file distributed with git-gud or https://github.com/benthayer/git-gud/blob/master/LICENSE.md for complete details.
```
benthayer: Can you mention specifically that this is the MIT license? It makes sense to keep everything else

git-gud.1
```diff
@@ -0,0 +1,68 @@
+.TH git-gud 1 "19 Feb 2020" "version 0.0"
```
benthayer: Git is written `GIT(1)` in its man page, maybe we should make this all caps too?

git-gud.1
e-leclercq: No, as this is supposed to be all lowercase.

e-leclercq: No, as this is supposed to be all lowercase.

git-gud.1
e-leclercq: Again, this is supposed to be lowercase. If you view the man page, you will see this.

git-gud.1
e-leclercq: Okay, I will make the change from `python\ \-m\ gitgud\ start` to `git gud init`, possibly moving the added change to a more relevant section.

git-gud.1
e-leclercq: I filed this under #145: "Add CONTRIBUTORS file to repository". Man pages are meant to be local manuals, so having a link in this section is a *faux pas*, which is why I suggested adding a contributing section. 

Also, note that the `AUTHOR` section of a man page is the person who wrote the man page. The `AUTHORS` section is reserved for both the person or persons who authored the man page **and** those who have contributed to the development of the program.

For now I am leaving this; however, upon resolution of issue #145, I will add that file to the man page and update the section to `AUTHORS` to reflect that change.

git-gud.1
e-leclercq: Ope, I must have accidentally held Shift while CapsLock was on. I will fix this.

git-gud.1
e-leclercq: This section contains the disclaimer, not the license itself; however, it refers to the LICENSE file you have distributed which is the MIT License.

git-gud.1
e-leclercq: Fixed

git-gud.1
e-leclercq: Fixed

git-gud.1
benthayer: Can I see a source for this?

I found [this](http://man7.org/linux/man-pages/man7/man-pages.7.html), which says the use of `AUTHORS` is strongly discouraged. When I run `man vim` and `man git`, the `AUTHOR` section 

Git man page:
```
AUTHORS
       Git was started by Linus Torvalds, and is currently maintained by Junio C Hamano. Numerous contributions have
       come from the Git mailing list <git@vger.kernel.org[5]>. http://www.openhub.net/p/git/contributors/summary
       gives you a more complete list of contributors.

       If you have a clone of git.git itself, the output of git-shortlog(1) and git-blame(1) can show you the authors
       for specific parts of the project.
```

Vim man page:
```
AUTHOR
       Most of Vim was made by Bram Moolenaar, with a lot of help from others.  See ":help credits" in Vim.
       Vim  is based on Stevie, worked on by: Tim Thompson, Tony Andrews and G.R. (Fred) Walter.  Although hardly any
       of the original code remains.
```

These seem to be similar usage, and given that use of the `AUTHORS` section is "strongly discouraged", I think the git approach of ```If you have a clone of git.git itself, the output of git-shortlog(1) and git-blame(1) can show you the authors for specific parts of the project.``` It also seems like `AUTHOR` and `AUTHORS` are the same section, just renamed depending on the number of authors, so it wouldn't make sense to have both sections.

Also, as this project is based on git, I want to follow their convention. It seems like the git repo contains the equivalent of the CONTRIBUTORS file and using the git log instead makes our lives easier.

So the bottom line would be to have that same quote from the git repository while also citing the other things I said.

I'd like to hear your thoughts.

git-gud.1
benthayer: I viewed the man page, and I see what you mean.

Instead, can you change the bolded "git-gud" to unbolded "Git Gud"? The official name is "Git Gud" and the command is "git gud" just like "Git"'s command is "git"

git-gud.1
```diff
@@ -0,0 +1,68 @@
+.TH GIT-GUD 1 "19 Feb 2020" "version 0.0"
+.SH NAME
+git-gud \- get good at git with git-gud
+.SH SYNOPSIS
+.B git gud
+[option1] [option2]
```
benthayer: I don't think this is actually the syntax we use. I'm thinking it might be better to take what git does:
```
SYNOPSIS
       git [--version] [--help] [-C <path>] [-c <name>=<value>]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p|--paginate|--no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           [--super-prefix=<path>]
           <command> [<args>]
```

And do this:

```
SYNOPSIS
       git gud [--version] <command> [<args>]
```
or something similar

git-gud.1
```diff
@@ -0,0 +1,68 @@
...
+.SH DESCRIPTION
+.B git-gud
+teaches you how to use Git with fun interactive tutorials.
+Use the following command to create a Git repository and start the first level:
+.B git\ gud\ init
```
benthayer: From the previous review:

> I looked at the man page for git, and it seems that all the git commands where you write git init etc are actually all their own commands and all have their own man page. Would it make more sense to actually convert git gud's subcommands into actual subcommands so things like git-gud-load are actually their own unique command and not just an argument to the git-gud command? That seems to be what git does with all of it's commands, and I'd love to stay in line with what Git's doing.

Can you please respond to this?

I'm thinking visually, this would look like `git-gud-init(1)` and I would have to create a new command entrypoint that would directly correspond with the `git-gud-init` command rather than a processed subcommand

git-gud.1
e-leclercq: This is true; however, the package name is git-gud. So, I figured that would go under name. For example, URxvt (which has a man page that is accessed with `man urxvt`) comes from the package `rxvt-unicode`, and the latter is what is under `NAME` in the man page. I think this makes things easier (at least, for myself it does); however, I can understand your point.

git-gud.1
e-leclercq: I like it. I did not know the exact syntax, so I will make the relevant changes right now.

git-gud.1
e-leclercq: Okay, that sounds good. I think that can be done after the PR though, as these would be separate man pages and this one is just for git-gud for now.

benthayer:
Can you push your updates?

e-leclercq:
I made the requested changes and just pushed them.

git-gud.1
```diff
@@ -0,0 +1,70 @@
...
+\fBinstructions\fP
+Displays instructions for current level.
+.TP
+\fBreset\fP
+Resets the current level.
+.TP
+\fBtest\fP
+Tests if current level is complete.
+.TP
+\fBload\ next\fP
```
benthayer: Move this by the other load commands.
There are also two more commands to add: `load prev` and `load previous`, which both load the previous commands

git-gud.1
```diff
@@ -0,0 +1,70 @@
+.TH GIT-GUD 1 "19 Feb 2020" "version 0.0"
+.SH NAME
+git-gud \- get good at git with git-gud
+.SH SYNOPSIS
+.B git gud
+[\-\-version] [\-\-help]  
+<command> [<args>]
+.SH DESCRIPTION
+.B Git Gud
```
benthayer: This is showing a weird space because of the bold. The git manpage doesn't have "Git" bolded, so I think it'd make sense to have this not bolded.

What do you think?

e-leclercq: This is not true. The bold text does not result in a spacing issue that is just the justification that man pages use. I did make an unrelated minor change to the spacing somewhere else though. Also, most man pages I have seen (and I read at least one man page an hour at this point) have the name of the program emboldened, including Vim and the terminal emulator I use.

git-gud.1
e-leclercq: I just update this and pushed the changes.

PR #148, opened 2020-02-21
Added `show tree` as an alternative command.

### LINKED ISSUE #117: Allow users to type in git gud show tree
It seems like users are typing this in anyways when they mean to type in git gud show-tree. We should let them

### LINKED ISSUE #177: GitHub/GitLab + Provider specific levels
We should have levels that specifically train you to use various features of git hosting providers.

One way to do this would be to have associated blog posts with images to help with the stuff in the UI and then git gud would act as a companion.

Things to use would be closing issues via commit messages, adding pull/merge requests, merging and various other things

sahansk2:
Fixes #117.

`git gud show tree` has been added as a valid command. I should note that there are two parsers which are doing work here: one that is made visible to the user when `git gud help` is run, (`_show_parser_visible`), and one which does the actual processing (`show_parser`).

It's a bit hackish of a solution, so do you have any ideas on how to improve this, @benthayer ?


benthayer:
I agree that this is a bit of a hacky solution. I think that's because it isn't meant to be done. I updated it to reflect how I think it should be, but I'm going to have to think about this before I actually merge. 

Looking at #152 , do you think we could somehow detect when someone enters git gud show tree and fail by telling them to use "git gud show-tree"? Feel free to overwrite my code, we can always go back to it.

benthayer:
After thinking about this, I think we can just replace show-tree with show tree. 

Would you mind making those updates?

sahansk2:
show-tree has been fully replaced.
I also modified the code a bit so that putting additional commands like `git gud show solution` should be easier.

benthayer:
Merged

PR #158, opened 2020-03-08
Added data_files to setup.py

e-leclercq:
I added a data_files line to your setup.py, so the man page gets installed with git-gud.


benthayer:
It turns out the issue we were having with this was just with `pip install -e .` When you remove the `-e`, things are fine.

PR #159, opened 2020-03-09
Expanded level API

benthayer:
- Added new methods to Level
 - Split up code from `util.py`
 - Put user messages in a centralized location
 - Added new intro level
 - Modified NamedList so the first skill is labeled 0 and the first level in a skill is labeled 1
 - Other various cleanup


benthayer:
Hey @sahansk2, can you take a look at this and tell me what you think?

There's a bunch of changes, but let me know if the overall design is understandable, and of course, if you have comments on anything else, that would be appreciated!

benthayer:
@karywjh I'm working on an expanded level API. You might not need it, but this includes an example for how to make a level using a subclass. Here's an example: `gitgud/skills/intro/__init__.py`

benthayer:
@wertydoo I'm partially developing this new API for you so you can simply pull from a bank of status messages when putting together the messages you want

PR #187, opened 2020-04-22
Added `git gud show solution`

### LINKED ISSUE #154: git gud show solution
People should have a way to get the solution if they really are stuck.

We'll probably want to confirm with a `y/n` prompt or a `-y` flag

sahansk2:
Fixes #154 
The flag used to confirm `git gud show solution` is `--confirm`.


benthayer:
Some notes:
 - Implementation works as expected

Please fix:
 - User messages should be defined in `skills/user_messages.py`
 - There should probably be a solution method defined as an attribute of `skills/level_builder.py:Level`

Please comment:
 - Text seems verbose
 - You added this command as `git gud solution`. It might make more sense that way. What do you think? `git gud show X` Can still be reserved for other commands in the future

sahansk2:
Personally, I like the format `git gud X` -- it feels cleaner, in my opinion.

Speaking of cleaner -- what do you think of putting all the messages for `handle_solution()` in the same function? I like it because when I look at `user_messages.py`, I can tell which function the messages belong to more easily.

gitgud/skills/user_messages.py
```diff
@@ -74,3 +74,15 @@ def all_levels_complete():
 def default_fail():
     print('Level not complete, keep trying. "git gud reset" to start from scratch.')
 
+@separated
```
benthayer, line 77: Is this something that needs to be separated? This decorator separates two blocks of text with one empty line. Are we ever using this with multiple blocks of texts (a block of text is equal to one other method with the `@separated` decorator)

gitgud/skills/user_messages.py
```diff
@@ -74,3 +74,15 @@ def all_levels_complete():
...
+def handle_solutions_messages(level, message_id):
```
benthayer, line 78: I think we need to split this up and give appropriate names. When you're calling it, you have no idea what you're going to get out of it.

A good example might be `no_solution_available`, `confirm_show_solution` or simply `show_solution`.
Whatever the name, it should be useful, but the method should only do one thing

gitgud/__main__.py
```diff
@@ -68,13 +69,13 @@ def __init__(self):
         ])
         
         show_description = "".join([
```
benthayer, line 71: Is there a reason we're not already just doing `"\n.join...`? Seems like it might be a better option

sahansk2: The reason that's there is because we've got more formatting in each individual line coming up for each subcommand. I was unsure about cramming together the escape characters with the actual content in the same line. If we switch to "`\n".join(...)`, we'd go from:
`"    tree", "\t", "Show the current state of the branching tree"`
to 
`"    tree\tShow the current state of the branching tree"`

If you'd like, @benthayer , I can switch it anyway.

benthayer: Yeah, that's how it's normally how we do it. I think it's better than having so many newlines

sahansk2:
I think it's worth mentioning that it's not immediately apparent that the levels created by `make_level.py` will break (pytest doesn't complain at all) -- it's just that `git gud solution` will say that there are no solutions available for the level.

Should I update make_level.py and the templates so that `git gud solution` works for it @benthayer? Or would you like to handle that?  

Sidenote: this is a pretty wack way to make a level that doesn't expose solutions at all to the user. Would we ever want such levels?

gitgud/skills/user_messages.py
```diff
@@ -76,3 +77,17 @@ def all_levels_complete():
...
     print('Level not complete, keep trying. "git gud reset" to start from scratch.')  # noqa: E501
+
+
+def handle_solution_confirmation(level):
+    print("Are you sure you want to view the solution for ", end="")
+    print('"{}": "{}"?'.format(level.name, level.skill.name))
+    print('If so, run `git gud solution` with --confirm.')
+
+
+def handle_solution_none_available():
```
benthayer, line 88: `no_solutions_available`

gitgud/__main__.py
```diff
@@ -340,6 +349,20 @@ def handle_test(self, args):
         level = self.file_operator.get_level()
         level.test(self.file_operator)
 
+    def handle_solution(self, args):
```
benthayer, line 352: This code should be in the `Level` class as the `solution` function. See `goal_str` and `goal`

benthayer: Also see `handle_goal`

gitgud/skills/rampup/_detaching/solution.txt
```diff
@@ -0,0 +1,4 @@
+# User would instead check out the commit hash using only one command
+git checkout bugFix
+git checkout @^
+git checkout @{1}
```
benthayer, line 4: The solution we're looking for is something like `git checkout asdf`. For this level, we should override the default `Level.solution()` method  and output the correct commit hash.

The test case should call a function to get the correct solution.

gitgud/skills/rampup/test_levels.py
```diff
@@ -4,30 +4,8 @@
...
 
-
 level_tests = [
-    (
-        # User would use something like "git checkout a1b2c3d4"
-        skill['detaching'], [
-            'git checkout bugFix',
-            'git checkout @^',
-            'git checkout @{1}',
-        ]
```
benthayer, line 15: Here, you're going to need to get the solution from the method that the level produces to get to the solution

gitgud/skills/parsing.py
```diff
@@ -1,6 +1,20 @@
 from copy import deepcopy
 
 
+def parse_solution(solution_path):
+    solution_commands = []
+
+    if not solution_path:
+        return solution_commands
```
benthayer, line 8: What's this for? When is `solution_path` going to be `None`?

gitgud/skills/parsing.py
```diff
@@ -1,6 +1,20 @@
 from copy import deepcopy
 
 
+def parse_solution(solution_path):
```
benthayer, line 4: This should be moved, similar to `BasicLevel.goal_str`.

gitgud/skills/parsing.py
```diff
@@ -1,6 +1,20 @@
...
+
+    if not solution_path:
+        return solution_commands
+
+    solution_data = solution_path.read_text()
+    for command in solution_data.split('\n'):
+        if not command or command.strip()[0] == '#':
+            continue
+        solution_commands.append(command)
+    return solution_commands
```
benthayer, line 15: I think this whole method could probably be `self.solution_path.read_text().split()`. See other comments

gitgud/skills/parsing.py
sahansk2: I implemented this for the case where `solutions.txt` doesn't exist as a file (because the level doesn't need solutions at all). The intro level, for example, doesn't have any solutions because the content of `test.spec` is the same as `setup.spec`.

Should we just operate under the assumption that all levels will have a `solution.txt` file? @benthayer

gitgud/skills/parsing.py
sahansk2: That'll work unless `solution.txt` doesn't exist for some reason. https://github.com/benthayer/git-gud/pull/187#discussion_r434074912

gitgud/skills/rampup/_detaching/solution.txt
sahansk2: Hm...well, this PR is already pretty long, so I think it's worth making a different issue/PR to autogenerate the solutions.txt file for that solution. I'm leaving this as is.

gitgud/skills/rampup/test_levels.py
sahansk2: See here: https://github.com/benthayer/git-gud/pull/187#discussion_r435473464

gitgud/skills/parsing.py
sahansk2: Also, doing `split()` would split each command into each keyword, e.g. `git gud commit` would be processed as `["git", "gud", "commit"]`, and the shell would try to run each of those as individual commands. So, I'm leaving the function as `split('\n')`.

gitgud/skills/parsing.py
sahansk2: I've just implemented it so that gitgud assumes there's always a `solutions.txt`.

benthayer:
@sahansk2 Can you rebase your recent changes related to completion tracking and put them on their own branch?

sahansk2:
@benthayer Not sure what you mean -- I didn't do anything regarding completion tracking here. The most recent commits did this, in order:
- 0982865: Got rid of the merge conflict markers and fixed a function call
- 36caf37: Added a `solutions.txt` template and added it to make_level.py's code 
- 04e56f1: Made a user message into a single print statement and added a #noqa line to satisfy flake8
- 0af906d: Implemented some of your suggestions about the `solutions.txt` parsing
- 767408b: Updated make_level.py's code because test_levels.py has the same code in every single skill now, because of the refactored changes

benthayer:
Lol woops. Got confused with the --confirm there 🤪

sahansk2:
Haha, no worries! By the way, I've dealt with most of the issues that you brought up, so feel free to review this. @benthayer

gitgud/skills/rampup/_detaching/solution.txt
benthayer: As long as we pass test cases, that's fine. I made a new issue in #215

gitgud/skills/parsing.py
benthayer: `split('\n')` is totally fine

make_level.py
```diff
@@ -120,40 +120,7 @@ def write_test(skill_path, level_name):
...
-
-        filedata = filedata.replace("\n".join([
-            "    )",
-            "]"
-        ]), replace)
-
-        with open(test_levels_path, 'w') as fp:
-            fp.write(filedata)
-            print('Created test case "{}" in {}'.format(level_name, test_levels_path))  # noqa: E501
+    print('Created test case "{}" in {}'.format(level_name, test_levels_path))  # noqa: E501
```
benthayer, line 123: Can we make the output conditioned on whether the file is currently present?
Something along the lines of:
"Creating test_levels.py for skill X"
"test_levels.py already exists"

PR #190, opened 2020-04-28
Fixed KeyError on Level 0

### LINKED ISSUE #182: KeyError `git gud load 1 0`
This is expected, but we should handle this cleanly so the user doesn't see this

sahansk2:
Fixes #182


benthayer:
Great fix, very simple

Issue #32, opened 2019-10-18
Track which levels have been completed


benthayer:
Since users can complete challenges out of order, it might be fun to save which challenges have been beaten. It would probably make the most sense to store this information in a json file in the `.git/gud` directory that is created after we run `git gud start`. It might be good to name it `progress.json` and store three values so we know whether a level has been attempted yet

The structure of the json in `progress.json` would be like this
```
{
'intro commits': 'complete',
'intro branching': 'incomplete',
'intro merging': 'unviewed',
...
} 
```

We may want to add more later on, but this is a good start and will allow us to just use the full names of the challenges when determining whether they're complete.

When loading levels, we can output something like "This level has already been complete" or "This level has not yet been completed", probably something more fun.

If someone wants to check their level progress, they can type in `git gud progress` or it might make sense to show the progress when using `git gud levels`. I'll let whoever works on this issue decide what to do there.
If we use `git gud progress`, we'll need to wait for #31 to be complete since we're going to be using the same command.

We will also have to be conscious of new levels being added, so if a key is not in that file, we should add it. 

Relevant method is `handle_progress` in `gitgud/__main__.py`

benthayer:
@wertydoo Do you want to tackle this?

benthayer:
@sahansk2 Can you comment here so I can assign the issue to you

sahansk2:
@benthayer Zoop :sunglasses:

PR #192, opened 2020-04-28
Added confirmation to git gud load next

### LINKED ISSUE #183: Confirm `git gud load next` if level is not complete
We can either do this via `git gud load next --force` (with the force option anywhere)

or we can do `confirm [y/n]`

Both should have relevant messages such as `Level not complete`

sahansk2:
Fixes #183

### DIFF
--- a/gitgud/__main__.py
+++ b/gitgud/__main__.py
@@ -9,6 +9,7 @@ from gitgud.operations import Operator, get_operator
 from gitgud.skills import all_skills
 from gitgud.skills.user_messages import all_levels_complete
 from gitgud.skills.user_messages import show_tree
+from gitgud.skills.user_messages import handle_load_confirm
 from gitgud.skills.user_messages import handle_solution_confirmation
 from gitgud.skills.user_messages import show_skill_tree
 
@@ -164,6 +165,7 @@ class GitGud:
                 metavar='level',
                 nargs='?',
                 help='Level to load')
+        load_parser.add_argument('--force', action="store_true")
 
         skills_parser = self.subparsers.add_parser(
                 'skills',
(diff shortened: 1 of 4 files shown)


gitgud/__main__.py
```diff
@@ -328,13 +330,21 @@ def handle_load(self, args):
...
         if args.skill_name:
             if args.skill_name.lower() in {"next", "prev", "previous"}:
                 query = args.skill_name.lower()
                 level = self.file_operator.get_level()
                                 
                 if query == "next":
-                    level_to_load = level.next_level
+                    current_level_complete = level._test(self.file_operator)
+                    if not args.force and not current_level_complete:
+                        print("It doesn't look like you've completed this level yet!")
```
benthayer, line 342: Can you put this output into `skills/user_messages.py`

benthayer:
The current output:
```
It doesn't look like you've completed this level yet!
Are you sure you want to go to the next level? If so, run `git gud load next` again with --force.
```
 seems a bit verbose, do you think you could come up with something shorter?

If you want to discuss, feel free to message me

sahansk2:
Wait a second. I think we should wait to merge this until we have a framework to keep track of what levels users have completed as they go through git gud. Otherwise, if they're just going through the levels that they've completed in order again (maybe to look at the various trees or whatever, who knows), they'll be reminded that they haven't completed the level yet, which is not necessarily true. What do you think? @benthayer

benthayer:
I think that's a good idea. Maybe this is a good time to start keeping track of which levels have been completed. 

I was holding off because I knew that the level names would change, but maybe it'd be worth giving each level a uuid so it becomes robust to name changes. Or we just wait.. What do you think?

benthayer:
@sahansk2 see #32 for tracking completed levels. It is currently our oldest issue :+1:

benthayer:
We are waiting on #241

gitgud/__main__.py
```diff
@@ -406,6 +420,10 @@ def handle_load(self, args):
                 level = get_operator().get_level()
 
                 if args.skill_name == "next":
+                    current_level_complete = level.get_progress() == "complete"
```
benthayer, line 423: Merge master and use `level.has_ever_been_completed`

gitgud/__main__.py
```diff
@@ -175,9 +177,21 @@ def __init__(self):
...
-        levels_parser.add_argument('skill_name', metavar='skill', nargs='?')
-        levels_parser.add_argument('--all', dest='opt_all', action='store_true', help="Prints all available skills with levels.")  # noqa: E501
-        levels_parser.add_argument('--short', dest='opt_short', action='store_true', help="Prints with the short name of skills/levels usable with `git gud load`.")  # noqa: E501
+
+        levels_parser.add_argument(
+                'skill_name',
+                metavar='skill',
+                nargs='?')
+        levels_parser.add_argument(
+                '--all',
```
benthayer, line 186: When merging master, you're going to need to add the `-a` option here

benthayer: Done

PR #223, opened 2020-06-10
Updates to contributing.md

### LINKED ISSUE #222: Update CONTRIBUTING.md to reflect current state of Git Gud
Since #187 was merged, the process of developing a level has changed slightly, so it would be nice to update [CONTRIBUTING.md](https://github.com/benthayer/git-gud/blob/master/CONTRIBUTING.md) with relevant information.
It's possible that there is additional information that is slightly outdated.

benthayer:
Fixes #222

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ git gud init # Initializes a git repository and lets us start using the git gud 
 Once you have the code downloaded, I would recommend having two terminals open at all times, one in the test directory you just created, and one in the git-gud directory. 
 This allows you to both work with your code and test your code at the same time.
 
-## Testing (Pytest)
+## Testing (pytest)
 Our tests make sure that each of the levels can be properly solved and that various other things work as expected.
 If you're fixing a bug, you should probably add a test that fails if that bug is present.
 #### To run tests:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -17,10 +17,24 @@ git gud init # Initializes a git repository and lets us start using the git gud
...
+Please make sure you are running tests for the gitgud subdirectory
+```
+pytest gitgud
+```
+or
+```
+cd gitgud
+pytest .
+```
+Pytest will search for files that start with `test_` and will run methods starting with `test_` as tests.The search ir run for all subdirectories, so all tests defined will automatically be found and don't need to be registered.
```
sahansk2, line 33: > tests.The search ir run

Typo?

sahansk2:
> @sahansk2 What do you think of these updates?

Updates look good -- I opened a PR to fix some typos and to explicitly mention `solution.txt`.

> Also, in general, what are your thoughts on making a discourse for potential contributors or users to ask questions instead of the potentially more uncomfortable way of submitting issues on GitHub?

If by discourse you mean something like a Discord, then I'm all for the idea! I know of various projects which have a Discord where people could drop comments/request help/etc.
Here are two such GitHub projects:
* https://github.com/FlagBrew/Checkpoint
* https://github.com/PSI-Rockin/Corgi3DS

I'm also open to the possibility of it not being on Discord specifically.

PR #227, opened 2020-07-01
Created remove_commit Level in Skill Rework

sahansk2:

PR #228, opened 2020-07-01
Added solution.user.txt to show user-oriented solutions.

sahansk2:
This adds compatibility for a new file in levels, `solution.user.txt`, where `git gud solution --confirm` will output the contents of this file instead of the `solution.txt`.
This is useful for when we have programmatic solutions for the level that we don't want the user to actually use as a solution. 

`makelevel.py` has not been updated because most levels' solutions in `solution.txt` are what we want the user to use, but there are still levels that need `solution.user.txt`.

### DIFF
--- a/gitgud/skills/intro/_welcome/solution.user.txt
+++ b/gitgud/skills/intro/_welcome/solution.user.txt
@@ -1,0 +1,1 @@
+git gud test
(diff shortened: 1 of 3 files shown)


sahansk2:
Closed and cherry-picked over to #227

PR #236, opened 2020-07-15
Refactors with pathlib

sahansk2:
This makes stuff more readable and also reduces the number of imports.


sahansk2:
Intro: Welcome doesn't have `explanation.txt` or `goal.txt`, and they're not overridden, so tests are failing :/

gitgud/__main__.py
```diff
@@ -317,21 +316,21 @@ def handle_init(self, args):
                 passargs = ''
 
             with open(path, 'w+') as hook_file:
-                hook_file.write('#!/bin/bash' + os.linesep)
+                hook_file.write('#!/bin/bash' + '\n')
```
sahansk2, line 319: By what the [docs](https://docs.python.org/3/library/os.html#os.linesep) say, we shouldn't use os.linesep for this purpose.

PR #240, opened 2020-07-22
Added test to assert False if the level is incomplete.

### LINKED ISSUE #216: Don't confirm `git gud solution` if level has ever been completed
Once #32 and #187 are done, we should allow users to run `git gud solution` without needing to confirm if they have already completed the level at any point. Code would go in  `Level.has_ever_been_completed()`

### LINKED ISSUE #233: Add test to make sure levels only pass if completed
If a user hasn't completed a level, but `_test` returns true, that is an issue. Right now pytest only checks to make sure `_test` returns true after a user has completed a level

We would need to add a specific check for any levels where nothing needs to be done, like `intro-welcome`

sahansk2:
Fixes #233
I understand why the `rework/irebase` level is failing, since you don't actually have to delete the `test` branch for the level to pass.
What I don't understand is why the `intro/config` level is failing tests.


sahansk2:
We should implement a flag for checking strict branch equivalence for `rework/irebase`.

benthayer:
Just writing down my plan. If you think you can do this, let me know and I can hold off and let you go for this.

I'd probably start by refactoring the code in `BasicLevel._test` such that the first half of the code goes into a `get_named_tree` method, and then I'd add a flag to `test_ancestry(skill, test)` like `test_ancestry(skill, test, no_extra_branches=False)`. To finish the job I'd override `_test` for `irebase` with `no_extra_branches=True`

Alternatively, we could change the solution so that it doesn't delete the branch. Do we care about the extra branch? We could even be strict about branches for every level

sahansk2:
I think that we should care about the extra branch.

For the solutions that we want to teach the user about, there exist other, non-optimal solutions that do the same thing; the computer-generated solutions are such non-optimal solutions. If other branches that aren't in the spec exist at the time of running `git gud test`, then that means that a "wrong" solution was used.

In my opinion, there shouldn't be random artifacts left by an attempt to restructure a git-repository -- at the very least, we should hone the user to clean up after themselves by enforcing strict branch checking.

What do you think, @benthayer ? It's possible that I'm being more strict than I need to be.

benthayer:
How would you feel about letting the user pass the level but telling them that they have a solution that's only partially collect? I think that's something we would want anyways. Can you think of any times we would want a user to pass with an imperfect solution?

Some things that come to my mind:
 
 - Extra branches
 - Extra tags
 - Typos/slightly incorrect branch names
 - They committed an extra file
 - Small typos in files that they were supposed to modify
 - Situations peculiar to an individual level

A system to provide error messages could be useful to help guide the user in the right direction, regardless of if the level is marked completed. 

How do you feel about partial solutions? Can you think of other situations where a user might partially solve a level?

benthayer:
Just updated this. Checks pass. Do you have anything else to add before I review this?

sahansk2:
@benthayer Nope, nothing that I can think of. This looks good!

PR #241, opened 2020-07-26
Track level progress 

### LINKED ISSUE #32: Track which levels have been completed

sahansk2:
Fixes #32 

I changed the format to: 
```
{  "intro"  : { "welcome" : "unviewed", "config" : "unviewed", "init" : "unviewed" },
   "basics" : { ... }, 
   ...
}
```
since instead of having `progress_data[skill.name][level.name]` to get the progress ffor that level, we'd have to do something like `progress_data[skill.name + " " + level.name]`, which seems unnecessarily verbose. It's not any better with `' '.join()`. But if you'd like, @benthayer , I can change it back to what #32 mentions.


sahansk2:
Oop, didn't see https://github.com/benthayer/git-gud/issues/32#issuecomment-663885727 !
@wertydoo I haven't done really much of anything here, and I've got other issues to look at as well, so feel free to take this issue!

benthayer:
@sahansk2, this is worth reopening since you've already done the work and @wertydoo hasn't responded.

The only thing missing are wrappers around the methods you wrote to get/set the progress of each level. Ex: `mark_level_passed(level)` or `get_level_progress(level)`

It would also be good to add a test. All you need to do is initialize the file and then check to make sure you can read an expected level.

sahansk2:
I made a new file for testing, `gitgud/test_operator.py`, since there really wasn't a good place to put tests related to file operators.

Also, I set the `progress_data` fixture to function scope, since we wouldn't want failing tests due to stale data.

sahansk2:
Before I implement the functions into levels -- does the API look good? @benthayer

gitgud/test_operator.py
```diff
@@ -0,0 +1,45 @@
...
+
+@pytest.fixture(scope='function')
+def progress_data(file_operator, gg):
+    progress_data = file_operator.read_progress_file()
+    return progress_data
+
+
+def test_get_level_progress(file_operator, progress_data):
+    level = all_skills["1"]["1"]
+    assert file_operator.get_level_progress(level) in {"complete", "incomplete", "unvisited"}  # noqa: E501
```
benthayer, line 22: Can you split this to multiple lines? In general, I prefer not adding E501 exceptions except for user messages

benthayer:
Things are definitely good enough to put into the level API

sahansk2:
```
user@ubuntu:test$ git gud levels --all
All levels and skills:

0. Introduction (intro)
├──✪ 1. Welcome (welcome)
├──◌ 2. Configuring (config)
└──◌ 3. Initialization (init)
1. Basics (basics)
├──◎ 1. Introduction to Commits (committing)
├──◌ 2. Branching in Git (branching)
├──◌ 3. Merging in Git (merging)
└──◌ 4. Introduction to Rebasing (rebasing)
2. Rampup (rampup)
├──◌ 1. Detaching HEAD (detaching)
├──◌ 2. Relative References I: Using (^) (relrefs1)
├──◌ 3. Relative References II: Using (~) (relrefs2)
└──◌ 4. Reversing Changes in Git (reversing)
3. Rework (rework)
├──◎ 1. Cherrypicking (cherrypicking)
└──◌ 2. Interactive Rebase (irebase)
4. Mixed Bag (mixedbag)
└──◎ 1. One Commit (onecommit)
5. Extras (extras)
└──◌ 1. The Octopus Merge (octopus)
```
I'm now using `incomplete`=`BULLSEYE`, `complete`=`CIRCLED WHITE STAR`, and `unvisited`=`DOTTED CIRCLE`, since they're all the same size in the terminal, and it doesn't look as cluttered as it does with the ballot boxes in the terminal. Also, completion stars are great. I'm trying to find something to replace `BULLSEYE`, something that's the same size and that's circular.

I deleted a bunch of the `test_passed` methods in the Intro levels, since I wanted to mark levels as properly visited and they didn't seem to add any additional functionality.

benthayer:
Four statuses
- Unvisited
- Visited
- Partially Complete (partial)
- Complete

gitgud/skills/user_messages.py
```diff
@@ -110,8 +110,16 @@ def default_fail_no_reset():
 
 
 def show_skill_tree(items, show_human_names=True, show_code_names=True, expand_skills=False):  # noqa: E501
-    middle_entry_bookend = '├── '
-    last_entry_bookend = '└── '
+    from gitgud.operations import get_operator
```
benthayer, line 113: Try 
`import gitgud.operations`

below, `operations.get_operator`

gitgud/skills/user_messages.py
```diff
@@ -150,6 +158,11 @@ def display_entry(index, human_name, code_name, indent):
             else:
                 indent = middle_entry_bookend
 
+            file_operator = get_operator()
```
benthayer, line 161: `file_operator = operations.get_operator`

sahansk2:
@benthayer This is ready to be merged.

I changed the UTF-8 characters to ASCII characters for this PR, since there are other issues/PRs ( #216 and #183 ) that depend on the features implemented here and the usage of UTF-8 characters is purely cosmetic.

Once this is merged, we can (and should) open an issue calling for the replacement of the ASCII characters with proper UTF-8 symbols.

gitgud/operations.py
```diff
@@ -324,6 +326,51 @@ def get_current_tree(self):
...
+            self.update_progress_file(progress_data)
+
+    def mark_level_visited(self, level):
+        progress_data = self.read_progress_file()
+        progress_data[level.skill.name].update(
+            {level.name: "visited"}
+        )
+        self.update_progress_file(progress_data)
+
+    def mark_level_partial(self, level):
```
benthayer, line 367: All three of these methods should be implemented the same (mark_level_complete/visited/partial), and use the same underlying implementation. `mark_level(self, level, status)`

gitgud/operations.py
```diff
@@ -324,6 +326,51 @@ def get_current_tree(self):
...
+            json.dump(progress_data, progress_file)
+
+    def read_progress_file(self):
+        with open(self.progress_path) as progress_file:
+            return json.load(progress_file)
+
+    def update_progress_file(self, data):
+        progress_data = self.read_progress_file()
+        with open(self.progress_path, 'w') as progress_file:
+            progress_data.update(data)
```
benthayer, line 345: Same here, can we do this outside of the with statement?

gitgud/skills/level_builder.py
```diff
@@ -43,6 +43,7 @@ def _setup(self):
 
     def setup(self):
         self._setup()
+        self.mark_visited()
```
benthayer, line 46: We do not want to mark this as visited if it has already been completed. Can you add checks in this method to make sure we never overwrite a "complete" or "partial"? Likewise, "partial" should never be overwritten by anything but "complete"

gitgud/skills/level_builder.py
```diff
@@ -80,11 +81,24 @@ def test_passed(self):
...
+    def mark_visited(self):
+        file_operator = operations.get_operator()
+        level_progress = file_operator.get_level_progress(self)
+        if level_progress == "unvisited":
+            file_operator.mark_level_visited(self)
+
+    def get_progress(self):
+        file_operator = operations.get_operator()
+        return file_operator.get_level_progress(self)
+
```
benthayer, line 97: `mark_partial` should be here too

gitgud/skills/level_builder.py
```diff
@@ -80,11 +81,24 @@ def test_passed(self):
...
         default_fail()
 
+    def mark_complete(self):
+        file_operator = operations.get_operator()
+        file_operator.mark_level_complete(self)
+
+    def mark_visited(self):
+        file_operator = operations.get_operator()
+        level_progress = file_operator.get_level_progress(self)
+        if level_progress == "unvisited":
```
benthayer, line 91: Even though the net effect is the same, I think this would be better as a not equal to statement, since we don't care if it's unvisited, we care that it's not `complete` or `partial` because we don't want to override either of those

gitgud/test_operator.py
```diff
@@ -0,0 +1,54 @@
+import pytest
+
+from gitgud.skills import all_skills
+from gitgud.operations import get_operator
+
+
+@pytest.fixture
+def file_operator(gg):
```
benthayer, line 8: I don't think you need gg as a parameter here. Did we already talk about this?

gitgud/test_operator.py
```diff
@@ -0,0 +1,54 @@
...
+        "unvisited", "visited", "partial", "complete"
+    }
+
+
+def test_update_progress_file(file_operator, progress_data):
+    file_operator.read_progress_file()
+    level = all_skills["1"]["1"]
+    progress_data[level.skill.name].update({level.name: "value"})
+    file_operator.update_progress_file(progress_data)
+    assert file_operator.get_level_progress(level) == "value"
```
benthayer, line 32: Is this safe to do? Is there a possibility the dictionary will show "value" during other tests?

gitgud/test_operator.py
```diff
@@ -0,0 +1,54 @@
...
+    assert file_operator.get_level_progress(level) == "unvisited"
+    file_operator.mark_level_complete(level)
+    assert file_operator.get_level_progress(level) == "complete"
+
+
+def test_mark_level_partial(file_operator):
+    level = all_skills["1"]["1"]
+    assert file_operator.get_level_progress(level) == "unvisited"
+    file_operator.mark_level_partial(level)
+    assert file_operator.get_level_progress(level) == "partial"
```
benthayer, line 54: You should add tests to make sure that we cannot set a "completed" level to "visited"

gitgud/test_operator.py
sahansk2: No, not that I recall. This `gg` is here because I wanted to assure that Git Gud was initialized since `file_operator` would only ever be `None` if Git Gud was intialized.

gitgud/operations.py
sahansk2: I was thinking about doing it that way, but if we ever decided to (for one reason) change the structure of the JSON file, keeping thee functions separate would make it transparent to whoever was calling the function (e.g. changing the status from `"unvisited"` in the data to `"Unvisited"`).

Now that you've given the OK, though, I'll change this.

gitgud/test_operator.py
sahansk2: Yep, this is 100% safe to do. `value` won't appear in other tests because the fixture is destroyed at the end of the test function and is rebuilt for the other tests, since it has function scope. (Actually, `pytest.fixture(scope='function')` is redundant, since [it's the default](https://docs.pytest.org/en/stable/fixture.html#fixture-scopes) -- I will fix that decorator). Here's a short [demonstration](https://gist.github.com/sahansk2/07d374a4e96432775be69f10efb70612) of what happens.

gitgud/skills/level_builder.py
sahansk2: So, you'd rather it be expressed as `if not level_progress in {"complete", "partial", "visited"}`? I always understood this function to mean "if the level is not yet visited (`unvisited`), mark the level as visited".

sahansk2:
> One question we should answer is whether we want Level to be handling the hierarchy of completion or if we want the operator to be handling the hierarchy of completion. I think it makes sense to have the operator handle it because then we can call `operator.mark_visited` without worrying that it'll overwrite anything. I want to hear your thoughts on this though.

This sounds fine to me. `Operator` should deal with the heirarchy.

gitgud/test_operator.py
benthayer: That's fine, feel free to leave as is.

gitgud/skills/level_builder.py
benthayer: As stated elsewhere, `Operator` should handle the completion hierarchy, so feel free to implement that in a single function that lays out all of the hierarchy.

`operator.mark_progress(level, progress)`

```
progress_dict = {"complete":3, "partial":2, "visited":1, "unvisited": 0}
if progress_dict[progress] > progress_dict[old_progress]:
    # mark it
```
This is probably more explicit than either way. Let me know if you think this could be improved on or if there are issues with it

sahansk2: Yeah, that's basically what I'm doing right now, except with lists (not yet pushed):
```
    def mark_level(self, level, status):
        progress_data = self.read_progress_file()
        heirarchy = [
            "unvisited", "visited", "partial", "complete"
        ]
        current_progress = self.get_level_progress(level)
        if heirarchy.index(status) > heirarchy.index(current_progress):
            progress_data[level.skill.name].update(
                {level.name: status}
            )
            self.update_progress_file(progress_data)
```

benthayer: Yes, that's fine

PR #242, opened 2020-07-26
Implements auto-generated asciinema.gif 

### LINKED ISSUE #189: Develop a framework to automatically generate the gif preview in the README
This is related to #138 .
We should not need to open issues to fix the same thing: updating the GIF preview in the README. If we need to update that preview, we should have some sort of a framework which can (1) effectively record the demo commands in the asciinema session by itself, and (2) automatically converts the asciinema output recording into a GIF file, such that if we ever need to update the gif preview, we can just push to the file which contains all of the updated commands.

sahansk2:
Fixes #189 

The plan is to implement this as a GitHub action using something like [Add and Commit](https://github.com/marketplace/actions/add-commit) that triggers only on tagged versions. I'm not sure how to deal with the actual authentication of Add and Commit (since the repo's not mine) but I can set up everything else in the GitHub action so that by the very end, all that's changed is `preview.gif`.

The API for demo-magic.sh is [pretty simple](https://github.com/paxtonhare/demo-magic#functions). I'm faking a network install of git-gud and doing a local install instead, because I don't know when the PyPI deployment to will trigger. I've also set up the script so that it errors out if anything in the script doesn't work.

I've still yet to implement the GitHub action, but I'll push this now so that we could talk about what we want to show in the preview.


benthayer:
Looks pretty good. All of that makes sense. The only thing I would want is to make the commit in a pull request rather than directly on the master branch. You should be able to test this out on your own repo and have it automatically make a pull request to your own repository. I can repeat any steps to reproduce it on my repo.

sahansk2:
Closing for personal testing.

PR #243, opened 2020-07-27
Added git gud debug as a hidden command

sahansk2:
This isn't for any issue, but it might be nice to have this command in case things break, and we want to probe around.

We could test the most recently released version that's distributed by PyPI without having to search for the executable to import and test.

This might also be useful for the 1.0 release if we want the user to probe around for errors by themselves, in case there's an error that they have that we can't reproduce.

I omitted the `help` argument so that the output doesn't appear in `git gud -h`.

Note: `import readline` is necessary to support history in the console.

### DIFF
--- a/gitgud/__main__.py
+++ b/gitgud/__main__.py
@@ -84,6 +84,10 @@ class GitGud:
         )
 
         self.subparsers.add_parser(
+                'debug',
+                description='Debug git-gud with necessary imports set up.'
+        )
+        self.subparsers.add_parser(
                 'status',
                 help='Print out the name of the current level',
                 description='Print out the name of the current level')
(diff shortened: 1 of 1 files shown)


benthayer:
Could you provide more information on when this would be useful? Is there a specific scenario that inspired you to make this?

benthayer:
@sahansk2 Closing until further information is provided

sahansk2:
When this would be useful: 
> We could test the most recently released version that's distributed by PyPI without having to search for the executable to import and test.
> This might also be useful for the 1.0 release if we want the user to probe around for errors by themselves, in case there's an error that they have that we can't reproduce.

As for what inspired me to make this? It was the development of #241 so that I could reinitialize and modify the `progress.json` file without having to open a text editor to modify the values.

PR #252, opened 2020-07-30
Added automatic asciinema generation

### LINKED ISSUE #189: Develop a framework to automatically generate the gif preview in the README

sahansk2:
Fixes #189 

https://github.com/sahansk2/git-gud/actions/runs/188315928
Right now, everything's being uploaded in the job as an artifact, under `build`.

~~I don't know how/if you want to do the add and commit action, since that commit won't be tagged as a release by default. We could use some sort of pre-release tag `p*`, which triggers the gif job, which then uses `EndBug/add-and-commit@v4` to add the generated preview.gif under `v*`. Alternatively, we can just download the generated `preview.gif` in the artifacts and add/commit it ourselves, and then push the result as a tagged release. I bundled the `asciinema.json` file in `build` in case we want to make `preview.gif` ourselves. However, gif generation takes about 7 minutes to run. That being said, I'm leaving the rest to you, @benthayer , since this is now an issue of design choice with respect to the project repository.~~

> The only thing I would want is to make the commit in a pull request rather than directly on the master branch

Whoops, I didn't see this in #242 . I'll deal with this eventually.


benthayer:
@sahansk2 What's the status of this?

previewgif.sh
```diff
@@ -0,0 +1,57 @@
...
+pip install ${GITGUD_REPO}
+
+p "# We'll first make an empty directory for git gud."
+pe "mkdir my-git-gud"
+pe "cd my-git-gud"
+p "# Now, let's initialize!"
+pe "git gud init"
+wait || true
+
+p "# Here's a demonstration of some of the levels that we have in store for you!"
```
benthayer, line 34: This line ends up overflowing, is there any way to either change the dimensions? If not, maybe we should shorten this line

previewgif.sh
```diff
@@ -0,0 +1,57 @@
...
+pe "git gud commit"
+pe "git gud show tree"
+
+pe "git gud test"
+pe "git gud load next"
+pe "git branch bugFix"
+pe "git checkout bugFix"
+pe "git gud commit"
+pe "git gud test"
+pe "git gud levels --all"
```
benthayer, line 52: How long do we stay on this screen? It would be cool to give the users a chance to see the output here

.github/workflows/generate-preview.yml
```diff
@@ -0,0 +1,93 @@
...
+    - uses: actions/setup-node@v1
+      with:
+        node-version: '14'
+
+    - name: Set up packager-managed dependencies for gif build
+      # demo-magic.sh requires [pv]
+      run: |
+        sudo apt-get update
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
```
benthayer, line 34: Are these necessary?

sahansk2:
This was exceedingly difficult to develop. Right now, it's using `git describe --tags` for the name of the branch, *and* it's using fetch-depth zero for `checkout@v2` action, so if there's any reason the version tag isn't recognized, it'll default to what `git describe --tags` would show. But that [should] never happen anyway.

@benthayer , this is ready to be merged.

.github/workflows/generate-preview.yml
```diff
@@ -0,0 +1,93 @@
...
+      run: |
+        mkdir /tmp/build
+        cp ./asciicast.json /tmp/build
+        cp ./preview.gif /tmp/build
+
+    - name: Upload asciicast.json
+      uses: actions/upload-artifact@v2
+      with:
+        name: asciicast.json
+        path: /tmp/build/asciicast.json
```
benthayer, line 67: Why do we upload this one?

sahansk2: This is uploaded as an artifact and is not ever committed to the repository. *Ever.*

This is useful if we ever decide that we want the asciicast.json file for inspection, testing alternatives to asciicast2gif, and more.

.github/workflows/generate-preview.yml
sahansk2: Yes, because `pip install` breaks without them, AFAIK. It's been a while since I touched that portion of code.

previewgif.sh
benthayer: @sahansk2 This is the main thing preventing a merge right now

sahansk2: Ehhh, not sure. Pretty sure that you can't change it as it is, and that you'd have to use `tmux` or something like that, and record the `tmux` session, and, well....it's far easier to just split it into multiple lines.

`# Here are just some of the levels that you can try out!` is shorter.

previewgif.sh
sahansk2: Whatever the default wait time is between commands. I can fix that, though.

previewgif.sh
benthayer: As long as it fits and makes sense, feel free to put whatever text you want here

previewgif.sh
benthayer: Let's just add maybe 10 seconds. Whatever feels natural enough for someone to be able to scan the whole thing and read a few of the level names

benthayer:
How do we view the new gif?

sahansk2:
Well, since this branch is on my own repository, you'd have to check the status of the action here: 
https://github.com/sahansk2/git-gud/runs/982970345?check_suite_focus=true

I think that I'm going to upload the `asciicast.json` file as soon as it's done generating, so that we can download and view it locally, since gif generation takes pretty long.

sahansk2:
@benthayer Could you check this out?

Issue #129, opened 2019-12-31
Force users to update git


parkmichael777:
We would like to take advantage of the foremost features of git; therefore, we should force users to install the latest version of git.

Some notable features include:
- git switch: allows you to change branches (similar to git checkout except ONLY for branches)
- latest git hooks such as pre-merge-commit.

The versions of git are backwards compatible so this shouldn't cause problems for users of older versions. Still, here are some things we should note:
- We need to check the user's current version of git.
- Check if the latest version is available.
- Help the user update if needed.
- Problem: the binary installers are not available at the same time the latest version is released. Furthermore, the binary installers for different OS's are released at different times. Either we have to check whether the installer is available for the specific OS, or we have to walk the user through the source code compilation (which can be ehh).

Issue #218, opened 2020-06-07
Level: `git restore`


benthayer:
Command is used to get rid of changes you've made

sahansk2:
`git restore` isn't available in the official Ubuntu repos as of 06-29-2020, since the most recent version by `git --version` is `2.17.1`. 
It might be worth trying to tackle #129 along with this level.

benthayer:
We can set something up to block certain levels based on someone's version of git.

For CI, we can install a minimum version of git for us to test with and then run specific tests with the appropriate versions of git

PR #269, opened 2020-08-07
Added support to fetch git version

sahansk2:
A feature added in consideration of #129 .

[`version_info`](https://gitpython.readthedocs.io/en/stable/reference.html#git.cmd.Git.version_info) returns the active version of `git` as a tuple.

This doesn't fix issue #129 , but it is certainly useful for #129 's resolution.

### DIFF
--- a/gitgud/operations.py
+++ b/gitgud/operations.py
@@ -4,7 +4,7 @@ import datetime as dt
 import email.utils
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


benthayer:
This is actually more relevant for things like #218, where we don't update, but where we can skip over the level if the user doesn't have the right version of git. Thanks for doing this

PR #270, opened 2020-08-07
Force init

### LINKED ISSUE #122: Warn user when force initializing git gud
We probably shouldn't be deleting people's files, so it'd be a good idea to create a commit that saves them and then just dereference that commit. We can let git save the commit object, but we could also create our own copy of the backup in the `.git/gud/backups` folder or something. We can find the commit hash using `git gud backups` and restore using `git checkout <hash>`

benthayer:
Fixes #122

### DIFF
--- a/gitgud/__main__.py
+++ b/gitgud/__main__.py
@@ -1,11 +1,7 @@
-import sys
 import webbrowser
 from pathlib import Path
 
 import argparse
-
-from git import Repo
-from git.exc import InvalidGitRepositoryError
 
 import gitgud
 from gitgud import operations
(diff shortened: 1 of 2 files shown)


benthayer:
@sahansk2 Is the the desired behavior?

sahansk2:
@benthayer Yep, this looks good!

This doesn't have to be in this PR, but do you think that we could have a feature where upon force initializing with `--prettyplease`, all the files in the current directory are committed to a bare repository in `.git/gud`, so that the user could run `git gud restore-backup` and restore their files?

benthayer:
First, I think copying all the files is fine, we don't need to use git to archive files.

Beyond that, there may be some issues. If we keep archiving things. then we might end up with many more files than we'd expect. There are a couple things we could do.

I think one of the best ways to make sure that there aren't too many archives is to make an archive folder called `.git/gud/archives` and create a `.git/gud/archives/{timestamp}` folder with all the files in there. The only thing not copied would be `.git/gud/archives`, which would remain in place each time things are archived/deleted.

 If we want to go back, then we can copy that folder, or make a git gud script to copy it, which I think would be overkill (but would definitely expedite the process) Overall, I think it's possible, but I don't see us using that much if  `--prettyplease` does its job, but maybe with this improved functionality, we'd actually be more likely to use it and make use of it somehow

PR #280, opened 2020-08-11
Update add branch with master code, valid Python, flake8 fixes

sahansk2:
Tests are failing because there isn't an `explanation.txt` for level `changes/onexone`, and `_test()` hasn't been implemented yet either.


benthayer:
Noting that this was closed so it could be merged into master

PR #285, opened 2020-08-14
Ignore branch capitalization for branches created by user

### LINKED ISSUE #139: Ignore capitalization for user-created branch names
When running `git gud test`, the capitalization of the branch names should be disregarded. For example, if on the `basics-branching` level, the user is prompted to create the branch `bugFix` in one of the steps. If the user creates the branch `BugFix` instead of `bugFix` but does all the other necessary steps correctly, we should accept the solution and deem it solved, instead of rejecting it outright.

benthayer:
This works by calculating which branches are created by the user and then setting them all to lowercase.
Fixes #139

### DIFF
--- a/gitgud/skills/level_builder.py
+++ b/gitgud/skills/level_builder.py
@@ -2,6 +2,7 @@ from importlib_resources import files
 
 import yaml
 
+from .parsing import branches_to_lowercase
 from .parsing import test_ancestry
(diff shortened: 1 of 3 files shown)


benthayer:
@sahansk2 This is ready for review

sahansk2:
@benthayer Looks good from here!

PR #289, opened 2020-08-14
Add file_operator handled support for dealing with commit content

### LINKED ISSUE #288: Update Operator to get contents of commit files
We'd like to have a common method to get the contents of a file in a commit, something like
```
def commit_content(self, commit, filepath)
```
since this is pretty crucial for level development in #238.

`commit` could support taking a [Commit](https://gitpython.readthedocs.io/en/stable/tutorial.html#the-commit-object) object, or a string commit-ish.

`filepath` should support taking [Path](https://docs.python.org/3/library/pathlib.html) objects and string representations of paths.

sahansk2:
Fixes #288 

Right now, these tests are pretty hacky, and GitPython keeps marking untracked files that are added to the staging area as "deleted". I'm using [`iter_change_type`](https://gitpython.readthedocs.io/en/stable/reference.html#git.diff.DiffIndex) to get the changes. What are your thoughts, @benthayer ?


gitgud/operations.py
```diff
@@ -155,6 +157,54 @@ def commit(self, commit_message, parents, time_offset):
...
+            filepath = str(filepath)
+
+        commit_content = (commit.tree / filepath).data_stream.read().decode("ascii")  # noqa: E501
+        self._streamed_content[commit] = commit_content
+        return commit_content
+
+    def get_staging_area(self):
+        return self.get_tracked_changes("staging")
+
+    def get_working_area(self):
```
benthayer, line 180: Working directory

gitgud/operations.py
```diff
@@ -155,6 +157,54 @@ def commit(self, commit_message, parents, time_offset):
...
 
+    def file_in_commit(self, commit, filepath):
+        commit = self.repo.commit(commit)
+        return filepath in commit.tree
+
+    def get_commit_content(self, commit, filepath):
+        if commit in self._streamed_content:
+            return self._streamed_content[commit]
+
+        commit = self.repo.commit(commit)
```
benthayer, line 168: These two statements are out of order, `_streamed_content` may have keys of type either `Commit` or `string`. You should make sure it's the type you want first. (`string` is probably better)

gitgud/operations.py
```diff
@@ -155,6 +157,54 @@ def commit(self, commit_message, parents, time_offset):
                 skip_hooks=True)
         return commit_obj
 
+    def file_in_commit(self, commit, filepath):
```
benthayer, line 160: We should be consistent with accepting `Commit`s or `string`s. I think we should accept both in all cases

gitgud/operations.py
```diff
@@ -155,6 +157,54 @@ def commit(self, commit_message, parents, time_offset):
...
+        commit = self.repo.commit(commit)
+        return filepath in commit.tree
+
+    def get_commit_content(self, commit, filepath):
+        if commit in self._streamed_content:
+            return self._streamed_content[commit]
```
benthayer, line 166: This doesn't make sense, are we reading from the file or the commit? This dict should need `filepath`

gitgud/test_operator.py
```diff
@@ -64,3 +72,34 @@ def test_mark_level_hierarchy(file_operator, level):
...
+
+def test_file_in_commit(file_operator, gg, content_level):
+    gg.load_level(content_level)
+    assert not file_operator.file_in_commit("master", "welcome.txt")
+    assert file_operator.file_in_commit("master", "Welcome.txt")
+
+
+def test_get_commit_content(file_operator, gg, content_level):
+    gg.load_level(content_level)
+    assert "Welcome" in file_operator.get_commit_content("HEAD", "Welcome.txt")
```
benthayer, line 85: Add another test. If we're actually caching the result of this, we should check that it's in the dictionary, so we should check that the function returns the right result.

We should ideally also check to see if this method works with commits, and works when you call it once and call it again with a different file. 

To make this clear:
```
assert "Welcome" in op.gcc("HEAD", "Welcome.txt")
assert "Welcome" in op.gcc("HEAD", "Welcome.txt") # Since it's now cached
assert "Hello" in op.gcc("HEAD", "Hello.txt")
assert "Hello" in op.gcc(op.repo.head.branch, "Hello.txt")
assert "Hello" in op.gcc(op.repo.head.commit.hexsha, "Hello.txt")
```
There may be typos in that

gitgud/test_operator.py
```diff
@@ -22,6 +24,12 @@ def level():
     return all_skills["1"]["1"]
 
 
+@pytest.fixture
+def content_level():
+    from gitgud.skills import all_skills
+    return all_skills['intro']['welcome']
```
benthayer, line 30: I don't know that `welcome` is a content level, but it works for the tests

gitgud/operations.py
```diff
@@ -155,6 +157,54 @@ def commit(self, commit_message, parents, time_offset):
...
+
+    def get_staging_area(self):
+        return self.get_tracked_changes("staging")
+
+    def get_working_area(self):
+        changes = self.get_tracked_changes("working")
+        changes["added"] += self.repo.untracked_files
+        return changes
+
+    def get_tracked_changes(self, change_type):
```
benthayer, line 185: When might this method be used? I think we're more interested in the content of the working directory or index, not the changes. 

We're definitely interested in the content. For example imagine there are two files, 1.txt and 2.txt. 1.txt is already committed and 2.txt has been staged. I would want a method `get_staging_area_content()` that returns a dict with the content, like this: `{"1.txt": "Hello world", "2.txt": "Is this content?"}`

gitgud/operations.py
```diff
@@ -155,6 +157,54 @@ def commit(self, commit_message, parents, time_offset):
...
+        if change_type == "staging":
+            staged_files = self.repo.index.diff("HEAD")
+        elif change_type == "working":
+            staged_files = self.repo.index.diff(None)
+        else:
+            raise KeyError
```
benthayer, line 191: We shouldn't have this if statement, these should just be different methods

gitgud/operations.py
```diff
@@ -155,6 +157,54 @@ def commit(self, commit_message, parents, time_offset):
...
+        return self.get_tracked_changes("staging")
+
+    def get_working_area(self):
+        changes = self.get_tracked_changes("working")
+        changes["added"] += self.repo.untracked_files
+        return changes
+
+    def get_tracked_changes(self, change_type):
+        if change_type == "staging":
+            staged_files = self.repo.index.diff("HEAD")
```
benthayer, line 187: See `repo.index.entries`

gitgud/operations.py
```diff
@@ -155,6 +157,54 @@ def commit(self, commit_message, parents, time_offset):
...
+        elif change_type == "working":
+            staged_files = self.repo.index.diff(None)
```
benthayer, line 189: No need to use GitPython for this one

gitgud/test_operator.py
sahansk2: Yeah, it would be nice to have handcrafted levels for testing. I added a TODO there.

sahansk2:
@benthayer Yep, I understood all of your comments! I think this is ready to be reviewed: tests are passing and the behavior is as you requested it.

gitgud/operations.py
```diff
@@ -159,51 +159,53 @@ def commit(self, commit_message, parents, time_offset):
...
+        content = {}
+        for filepath in commit.tree:
+            content.update(
+                {filepath: self.get_commit_file_content(commit, filepath)}
+            )
+        return content
+
+    def get_staging_content(self):
+        content = {}
+        for (path, _stage), entry in self.repo.index.entries.items():
```
benthayer, line 190:
```
>>> list(op.repo.index.iter_blobs())[0]
(0, <git.Blob "788df9fc65e5d61b14b96c96fa9a0e2cd241b109">)
>>> list(op.repo.index.iter_blobs())[0][1]
<git.Blob "788df9fc65e5d61b14b96c96fa9a0e2cd241b109">
>>> list(op.repo.index.iter_blobs())[0][1].path
'1.txt'
>>> list(op.repo.index.iter_blobs())[0][1].data_stream.read()
b"Hello, I'm an auto-generated file!"
```

benthayer: We want to use `iter_blobs` instead. The first item in that tuple is the stage, as defined in the [3-way merge documentation ](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-read-tree.html)

gitgud/operations.py
```diff
@@ -159,51 +159,53 @@ def commit(self, commit_message, parents, time_offset):
...
+            index_blob = entry.to_blob(self.repo)
```
benthayer, line 191: Can you use the name `entry_blob` rather than `index_blob`

gitgud/operations.py
```diff
@@ -159,51 +159,53 @@ def commit(self, commit_message, parents, time_offset):
...
-        for change in change_key_map:
-            change_data[change] = [
-                filename.a_path for filename in
-                staged_files.iter_change_type(change_key_map[change])
-            ]
-        return change_data
+    def get_working_content(self):
+        content = {}
+        for path in self.path.glob('*'):
+            # TODO: Add support for directories
```
benthayer, line 200: Can you either add support here (preferred) or add an exception or some type of error handling?

gitgud/operations.py
```diff
@@ -159,51 +159,53 @@ def commit(self, commit_message, parents, time_offset):
...
-            "modified": "M",
-            "renamed": "R"
-        }
+    def get_commit_content(self, commit):
+        content = {}
+        for filepath in commit.tree:
+            content.update(
+                {filepath: self.get_commit_file_content(commit, filepath)}
+            )
+        return content
```
benthayer, line 186: This does not get the content of sub-trees (directories).

If I have a subdirectory, I can do something like this:
```
>>> repo.head.commit.tree.trees[0]
<git.Tree "1b7d3913a8f0d539997a1d7d21ae6181061fc7f9">
>>> repo.head.commit.tree.trees[0].blobs
[<git.Blob "c47b355949ca027c0f929e24343279458228a04f">]
```
[Trees](https://gitpython.readthedocs.io/en/stable/reference.html?highlight=to_blob#git.objects.tree.Tree) can have subtrees and we should handle that.

benthayer: `repo.head.commit.tree.traverse()` should do the trick. Be aware that this also returns trees, which you would likely want to skip.

gitgud/operations.py
```diff
@@ -159,51 +159,53 @@ def commit(self, commit_message, parents, time_offset):
...
 
-        change_data = {}
-        for change in change_key_map:
-            change_data[change] = [
-                filename.a_path for filename in
-                staged_files.iter_change_type(change_key_map[change])
-            ]
-        return change_data
+    def get_working_content(self):
```
benthayer, line 197: I know it's long, but `get_working_directory_content` would probably be better.

gitgud/test_operator.py
```diff
@@ -80,26 +83,42 @@ def test_file_in_commit(file_operator, gg, content_level):
...
 
-def test_get_commit_content(file_operator, gg, content_level):
+def test_get_commit_file_content(file_operator, gg, content_level):
     gg.load_level(content_level)
-    assert "Welcome" in file_operator.get_commit_content("HEAD", "Welcome.txt")
+    assert "Welcome" in file_operator.get_commit_file_content("HEAD", "Welcome.txt")  # noqa: E501
+    # Check caching of decoded blob data_stream
+    assert "Welcome" in file_operator.get_commit_file_content("HEAD", "Welcome.txt")  # noqa: E501
+    head_sha = file_operator.repo.head.commit.hexsha
+    assert "Welcome" in file_operator.get_commit_file_content(head_sha, "Welcome.txt")  # noqa: E501
```
benthayer, line 92: Add a test for directories

gitgud/test_operator.py
```diff
@@ -80,26 +83,42 @@ def test_file_in_commit(file_operator, gg, content_level):
...
+    staging_data = file_operator.get_staging_content()
+    working_data = file_operator.get_working_content()
+    commit_content = \
+        file_operator.get_commit_file_content("HEAD", "Welcome.txt")
+
+    assert staging_data["Welcome.txt"] == commit_content
 
-    assert "untracked.txt" in change_data["added"], change_data
+    assert "untracked.txt" in staging_data, staging_data
+    assert "untracked.txt" in working_data, working_data
```
benthayer, line 124: Add a test for directories

gitgud/test_operator.py
```diff
@@ -80,26 +83,42 @@ def test_file_in_commit(file_operator, gg, content_level):
...
 
-    change_data_working = file_operator.get_working_area()
+    assert "untracked.txt" in working_data
+    assert "untracked.txt" not in staging_data
 
-    assert "untracked.txt" in change_data_working["added"]
+    assert "Welcome.txt" in staging_data
+    assert staging_data["Welcome.txt"] == working_data["Welcome.txt"]
 
+    # Add untracked.txt to the index
```
benthayer, line 111: Here and below can probably be a separate test

gitgud/test_operator.py
```diff
@@ -27,6 +28,8 @@ def level():
 @pytest.fixture
 def content_level():
     from gitgud.skills import all_skills
+    # TODO: Make a proper testing level with details.yaml
+    # We're using intro/welcome because it specifies a details.yaml
```
benthayer, line 32: You can remove these lines, we'll get to that when/if it becomes an issue.

sahansk2:
Marking as draft, since this is currently incompatible with Windows:

https://github.com/sahansk2/git-gud/actions/runs/221208128

sahansk2:
Some comments on the errors...

```
         else:
>           return self.get_commit_content(commit_hash)[filepath]
E           KeyError: 'dir\\subdir\\subdirfile.txt'
```

I suppose pathlib's Path automatically resolves to the default operating system's separators. We might have to use only PosixPaths or something.

```
         # Test unmodified
        assert "Welcome.txt" in staging_data
>       assert staging_data["Welcome.txt"] == working_data["Welcome.txt"]
E       assert 'Welcome to G...!\r\n\r\n\r\n' == 'Welcome to G...it-gud!\n\n\n'
```
This probably has something to do with how Git works on Windows. I recall that when installing, you have to specify a checkout option, with the default being "Check out windows endings, commit unix endings". It might have something to do with that. Python might also be converting the `\r\n` to `\n` all by itself when it reads the file, which explains why `working_data` doesn't have the carriage returns, while `staging_area` makes use of the raw blob data in the index.

```
        for untracked_file in untracked_files:
>           assert untracked_file in working_data
E           AssertionError: assert 'dir/untracked.txt' in {'Welcome.txt': 'Welcome to Git Gud, a command line game designed to help you learn how to use the popular version con...rfile.txt': 'dir/subdir/subdirfile.txt content', 'dir\\subdir\\untracked.txt': 'dir/subdir/untracked.txt content', ...}
```
This probably has something to do with how the key is initialized. I think it's being directly copied over from an instance variable. 

```
 >           assert added_file in working_data
E           AssertionError: assert 'dir/added.txt' in {'added.txt': 'added.txt content', 'dir\\added.txt': 'dir/added.txt content', 'dir\\subdir\\added.txt': 'dir/subdir/added.txt content', 'dir\\subdir\\subdirfile.txt': 'dir/subdir/subdirfile.txt content', ...}
```
Same deal here.

That being said, I'm closing this so that I can rebase on a branch with windows testing via GitHub actions.

PR #301, opened 2020-08-24
Implement file_operator content

### LINKED ISSUE #288: Update Operator to get contents of commit files

sahansk2:
Fixes #288 

Replaces #289 with tests for all operating systems. 
https://github.com/sahansk2/git-gud/actions
CircleCI build error because there's no CircleCI `config.yml`

### DIFF
--- a/gitgud/test_operator.py
+++ b/gitgud/test_operator.py
@@ -124,21 +124,26 @@ def test_get_commit_content_contains(file_operator, content_level):
     assert Path("dir/dirfile.txt") in commit_content
 
 
-def test_get_working_staging_content_tracked(file_operator, content_level):
+def test_unmodified_content(file_operator, content_level):
     staging_data = file_operator.get_staging_content()
     working_data = file_operator.get_working_directory_content()
+
     # Test unmodified
     assert "Welcome.txt" in staging_data
     assert staging_data["Welcome.txt"] == working_data["Welcome.txt"]
+
+
+def test_modified_content(file_operator, content_level):
     # Test modified
     write_file("Welcome.txt")
     staging_data = file_operator.get_staging_content()
     working_data = file_operator.get_working_directory_content()
+
     assert "Welcome.txt" in staging_data
     assert staging_data["Welcome.txt"] != working_data["Welcome.txt"]
 
 
-def test_get_working_staging_content_untracked(file_operator, content_level):
+def test_untracked_content(file_operator, content_level):
     untracked_files = [
         "untracked.txt", "dir/untracked.txt", "dir/subdir/untracked.txt"
     ]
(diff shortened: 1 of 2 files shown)


sahansk2:
@benthayer , this is ready for review.

sahansk2:
@benthayer , the matrix strategy is properly set up, and I added further modifications so that we can see all occurring errors across all OS's, even if the other OS's failed, and even if the previous (relevant) steps failed, since we don't necessarily want to have multiple commits just to fix flake8, which is the test that's run last. 

See here: https://github.com/sahansk2/git-gud/runs/1024141817?check_suite_focus=true
I cancelled the workflow, but notice how even though pytest for ubuntu-latest failed, flake8 was run anyway.

So, this PR is ready for review again.

benthayer:
@sahansk2 Can you update this so that flake8 only runs once? For now, it's fine that the windows test uses bash, but we should fix that at some point. Also can you put all of the testing commits on another branch? I'd like to merge into master with that first, then get to reviewing this PR. Once we merge that branch into master, then it will be appropriate to merge master into this branch and do a review

sahansk2:
Caching of the commit content doesn't actually seem to be necessary. I've added a commit to show this is the case.

sahansk2:
@benthayer This is ready for review. Could you comment on what needs to be done in order for this to be merged?

gitgud/test_operator.py
```diff
@@ -22,6 +26,21 @@ def level():
...
+@pytest.fixture
+def content_level(gg):
+    from gitgud.skills import all_skills
+    gg.load_level(all_skills['intro']['welcome'])
+
+    write_file("root.txt")
+    os.mkdir("dir")
+    write_file("dir/dirfile.txt")
+    os.mkdir("dir/subdir")
+    write_file("dir/subdir/subdirfile.txt")
```
benthayer, line 38: Why do you add `root.txt` and `subdirfile.txt`? It seems like you don't use them

benthayer (commented):
I updated names and generally refactored everything. Most of the testing code is the same, but is just shifted around and put in their own methods. 

As for the implementation code, I mostly separated expressions and assigned them to intermediate variables to make things more clear. The biggest structural difference is the `DirectoryContent` object.

If you like the way things are, then I'm okay to merge

gitgud/test_operator.py
sahansk2: I've now added tests for get_commit_content using subdir*.

sahansk2:
This looks good, I just added subdirectory tests since they were unused.

benthayer:
Great, will check it out within a day or two

PR #298, opened 2020-08-22
Added windows testing as GitHub workflow

sahansk2:
Fixes #297

PR #304, opened 2020-08-25
Implement testing for all operating systems as GitHub Actions

### LINKED ISSUE #297: Set up workflow to run Git Gud tests in a Windows environment
While developing, we might accidentally make the code Unix-only, especially when dealing with the expression of paths and directories. However, they won't go noticed, since the only testing workflow we have runs on `ubuntu-latest`. A workflow (through GitHub actions or through CircleCI) that runs the current commands in a Windows container would remedy this issue.

sahansk2:
Fixes #297  , replaces #298 , commits taken from #301


sahansk2:
> @sahansk2 Can you update this so that flake8 only runs once? For now, it's fine that the windows test uses bash, but we should fix that at some point (can you add an issue?). Also can you put all of the testing commits on another branch? I'd like to merge into master with that first, then get to reviewing this PR. Once we merge that branch into master, then it will be appropriate to merge master into this branch and do a review

If we want to run flake8 only once, it's probably best that a separate workflow is created for that.

sahansk2:
@benthayer This is ready for review.

benthayer:
@sahansk2 It looks like there's something weird going on with Windows testing. Can you take a look? It's getting a `OSError('[WinError 6] The handle is invalid')`, but I'm not sure what that means. Git can be accessed from the (bash) command line, as seen by the new test I added, so there must be something else going on

sahansk2:
GitPython [released 3.1.8](https://pypi.org/project/GitPython/#history) on September 3rd, which apparently broke Windows compatibility. 

By the way: it might be better to lock GitPython at 3.1.7 for [reasons other than compatibility]((https://github.com/gitpython-developers/GitPython/commit/eb411ee92d30675a8d3d110f579692ea02949ccd)), for a while at least.

benthayer:
Interesting, go ahead and lock it in. We can deal with the upgrade later

sahansk2:
Yep, it's been locked in. @benthayer

PR #305, opened 2020-08-25
Level: First Commit

### LINKED ISSUE #246: Level: Basics 1 - "First Commit"
```
1) First Commit - Show that you can add/commit - Commits contain files (adapt from basics-committing)
```
Just add a file and commit that file.

Status shows your progress in a way somewhat like this.
```
 Created: Yes
 Added: No
 Committed: No
```

The level demonstrates the process and shows how files must be added before they are considered part of a commit

sahansk2:
Fixes #246, branched from 288-commit-content-2 branch (PR in #301). The API is stable, so I'm starting development on issues dependent on it.

### DIFF
--- a/gitgud/skills/newbasics/__init__.py
+++ b/gitgud/skills/newbasics/__init__.py
@@ -12,28 +12,23 @@ class FirstCommit(BasicLevel):
 
     @separated
     def status(self):
-        created, added, committed = self.get_solved_state()
+        created, added, committed = self.get_state()
         print("Created:", bool_to_word(created))
         print("Added:", bool_to_word(added))
         print("Committed:", bool_to_word(committed))
 
-    def get_solved_state(self):
-        created = False
-        added = False
-        committed = True
+    def get_state(self):
         file_operator = operations.get_operator()
-        created_files = file_operator.get_working_directory_content()
-        if created_files:
-            created = True
-        added_files = file_operator.get_staging_content()
-        if added_files:
-            added = True
-        committed = file_operator.repo.head.is_valid()
 
-        return bool(created), bool(added), bool(committed)
+        created = bool(file_operator.get_working_directory_content())
+        added = bool(file_operator.get_staging_content())
+        committed = file_operator.repo.head.is_valid() \
+            and bool(file_operator.repo.get_commit_content('HEAD'))
+
+        return created, added, committed
 
     def _test(self):
-        created, added, committed = self.get_solved_state()
+        created, added, committed = self.get_state()
         return created and added and committed
 
     def post_setup(self):
--- a/gitgud/operations.py
+++ b/gitgud/operations.py
@@ -32,6 +32,9 @@ class DirectoryContent:
             filepath = str(filepath.as_posix())
         return self.content[filepath]
 
+    def __bool__(self):
+        return bool(self.content)
+
 
 def normalize_commit_arg(commit_func):
     @wraps(commit_func)


benthayer:
#301 is merged, which you said was blocking this. Is this ready to be reviewed?

sahansk2:
@benthayer Yep, this is good for review. Sidenote, I wish there was a better way to check if a commit exists, but it seems like [a try-except is the only way.](https://github.com/gitpython-developers/GitPython/issues/721)

gitgud/skills/newbasics/_firstcommit/explanation.txt
```diff
@@ -0,0 +1,3 @@
+Use this file to explain to the users how to use the git feature that this level focuses on.
+>>>
+Separate blocks using ">>>", and be sure to include tips and examples as explanation aids.
```
benthayer, line 3: @sahansk2 This is incomplete

gitgud/skills/newbasics/_firstcommit/goal.txt
```diff
@@ -0,0 +1 @@
+In this file, write a short explanation of the level's goal (e.g. "Merge commits 1 and 2").
```
benthayer, line 1: This hasn't been modified either

gitgud/skills/newbasics/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class FirstCommit(BasicLevel):
+    def _setup(self):
+        file_operator = operations.get_operator()
+        file_operator.destroy_repo()
+        file_operator.use_repo()
+
+    def status(self):
+        firstcommit_status(*self._get_state())
```
benthayer, line 14: I think since this status message is not currently intended to be reused, we should have the code here

gitgud/skills/user_messages.py
```diff
@@ -177,3 +184,10 @@ def display_entry(index, human_name, code_name, indent):
                 code_name=item.name,
                 indent=indent
             )
+
+
+@separated
+def firstcommit_status(created, added, committed):
```
benthayer, line 190: See other comment, I think we can have this function be a part of the `FirstCommit` class rather than separate

gitgud/skills/newbasics/_firstcommit/solution.txt
```diff
@@ -0,0 +1 @@
+git gud commit
```
benthayer, line 1: When someone is solving this level, they will not run this command, they'll run something along the lines of
```
touch file
git add file
git commit -m "Initial Commit: Added file"
```

gitgud/skills/newbasics/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+            created = True
+        added_files = file_operator.get_staging_content()
+        if added_files:
+            added = True
+        try:
+            file_operator.repo.head.commit
+        except ValueError:
+            committed = False
+
+        return created, added, committed
```
benthayer, line 32: What do you think about doing `return bool(created_files), bool(added_files), committed`?

gitgud/skills/newbasics/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+        created_files = file_operator.get_working_directory_content()
+        if created_files:
+            created = True
+        added_files = file_operator.get_staging_content()
+        if added_files:
+            added = True
+        try:
+            file_operator.repo.head.commit
+        except ValueError:
+            committed = False
```
benthayer, line 30: `committed = file_operator.repo.head.is_valid()`

gitgud/skills/newbasics/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+
+        return created, added, committed
+
+    def _test(self):
+        return all(self._get_state())
```
benthayer, line 35: Even thought this works well and is the intended use, I think this might be more readable if we do
```
created, added, committed = self.get_state()
return created and added and committed
```
We don't need to do this (especially since it's not hard at all to look up what `get_state` returns, but what do you think?

gitgud/skills/newbasics/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+class FirstCommit(BasicLevel):
+    def _setup(self):
+        file_operator = operations.get_operator()
+        file_operator.destroy_repo()
+        file_operator.use_repo()
+
+    def status(self):
+        firstcommit_status(*self._get_state())
+
+    def _get_state(self):
```
benthayer, line 16: Since it would be totally fine for us to call this from another part of the code, I think we can call this method `get_state` without the underscore. @sahansk2 Do you think this conflicts with our other `_test`/`_setup` naming scheme?

benthayer (changes requested):
Seems like there can be a few improvements. Overall this is functional and is pretty good

gitgud/skills/newbasics/__init__.py
sahansk2: This is fine, but Python semantics would end up causing the return of the last item that was evaluated before the expression could be resolved. If we do the `bool` method you mentioned in the other comment, then this is fine.

gitgud/skills/newbasics/__init__.py
sahansk2: See [this comment](https://github.com/benthayer/git-gud/pull/305#discussion_r486033022).

PR #308, opened 2020-08-29
Implement logging and Git Gud Error

### LINKED ISSUE #279: Develop an error logging framework to hide errors from user unless prompted
Spawned from #197 .

https://github.com/benthayer/git-gud/issues/197#issuecomment-671157907: 
> Hmm...just to make sure, would you like some sort of error handling so that the error is hidden from the user unless they run `git gud error` explicitly (for debug purposes)? Something like [this, maybe?]
> > Oops! Git Gud ran into a problem. Please try your command again. If the problem persists, please run `git gud issues` to open our GitHub, and create an issue with the output of `git gud error` so that we can try to help you!

https://github.com/benthayer/git-gud/issues/197#issuecomment-671614510:
> I don't think we need to `pickle` the error, since the stack trace is only useful as text for us (and so, `pickle` seems overkill). We could always just output to two files, `.git/gud/error.txt` and `.git/gud/logs/hh:mm-m-d-y.txt` with `error.txt` holding the content of the most recent log. `git gud error` would print out `error.txt`.
> 
> I think we should try to make use of the [`logging`](https://docs.python.org/3/library/logging.html) module, since it has [support for writing to files with extra information](https://docs.python.org/3/howto/logging.html#logging-to-a-file).

sahansk2:
Fixes #279

### DIFF
--- a/gitgud/__main__.py
+++ b/gitgud/__main__.py
@@ -5,6 +5,7 @@ import sys
 import argparse
 
 import gitgud
+from gitgud import logging
 from gitgud.operations import Operator, get_operator
 from gitgud.skills import all_skills
 from gitgud.skills.user_messages import all_levels_complete
@@ -13,6 +14,7 @@ from gitgud.skills.user_messages import handle_load_confirm
 from gitgud.skills.user_messages import handle_solution_confirmation
 from gitgud.skills.user_messages import show_skill_tree
 
+log=logging.getLogger(__name__)
 
 class InitializationError(Exception):
     pass
@@ -301,7 +303,7 @@ class GitGud:
                     print('{} will be left as is.'.format(file_operator.gg_path))  # noqa: E501
                 return
             else:
-                print('Force initializing Git Gud.')
+                log.info('Force initializing Git Gud.')
         elif len(list(Path.cwd().iterdir())) != 0:
             if not (args.force and args.prettyplease):
                 print('Current directory is nonempty. Initializing will delete all files.')  # noqa: E501
@@ -325,7 +327,7 @@ class GitGud:
                 level.status()
             except KeyError:
                 level_name = get_operator().read_level_file()
-                print('Currently on unregistered level: "{}"'
+                log.error('Currently on unregistered level: "{}"'
                       .format(level_name))
         else:
             print("Git Gud not initialized.")
--- a/gitgud/__init__.py
+++ b/gitgud/__init__.py
@@ -2,11 +2,15 @@ from importlib_resources import files
 
 import sys
 import subprocess
+import logging.config
 
 from git import Actor
 
 actor = Actor("Git Gud", "git-gud@example.com")
 actor_string = "Git Gud <git-gud@example.com>"
+
+
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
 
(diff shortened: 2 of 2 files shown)


gitgud/__main__.py
```diff
@@ -13,6 +14,7 @@
 from gitgud.skills.user_messages import handle_solution_confirmation
 from gitgud.skills.user_messages import show_skill_tree
 
+log=logging.getLogger(__name__)
```
benthayer, line 17: There should be spaces for assignments: `log = ...`

sahansk2:
This is very much not ready for review. I will undraft this PR when it is.

benthayer:
@sahansk2 Does this actually solve #279? There are two specific things we wanted to log, can you add them? This is simple enough to get us started and hopefully to get into the habit of logging errors using the logger.

On another note, do you think we should convert all print statements to use the logger?

benthayer:
Also, can you merge master so you can get the GitHub Actions

sahansk2:
This will solve #279, but it doesn't right now. I wanted to make a draft PR ahead of time so that we could double check on progress. 

The biggest issues that I have right now are basically:
* How do we set up a single logger to simultaneously log to two different files?
* How do we automatically log exceptions without wrapping everything in a try-except block? I'm hesitant of doing this simply because it might obscure some exceptions, like how a KeyError in a level's status method causes Git Gud to raise an unrelated exception ("Current level doesn't exist").

We shouldn't convert everything to logging (see [here](https://docs.python.org/3/howto/logging.html#when-to-use-logging)), but we could definitely convert some print messages to logging (e.g. `Current level doesn't exist.`)

benthayer:
@sahansk2  Does this help? https://stackoverflow.com/questions/11232230/logging-to-two-files-with-different-settings

benthayer:
I think for logging, we can take a simple first approach of simply logging things whenever we think it's necessary, so if there's an exception, include a line with `logger.info(information)`

sahansk2:
> @sahansk2 Does this help? https://stackoverflow.com/questions/11232230/logging-to-two-files-with-different-settings

Unfortunately, not really. It would be a bit of a hassle on our end to always output to two loggers, one `error.txt` and one `YYYY-MM-DD-HH-mm.txt`. 

I just thought of this, though: what do you think about having a wrapper function, `gitgud_logger(*args, **kwargs)` that sends `args` and `kwargs` to two logger objects simultaneously? That would make logging much easier for us, and we can also use the two-file approach with `error.txt` and the time-specific file.

benthayer:
I see what you mean. What if when users had errors, we just logged it all to one file and had them send that file to us

sahansk2:
> I see what you mean. What if when users had errors, we just logged it all to one file and had them send that file to us?

That could work, but it might be difficult from a development perspective to sift through it and see the most important errors. Having an `error.txt` with the most recent issue is the best approach for now. A bug tracking system would be nice to have; we'd probably want it to be opt-in (for automatic), or on a case-by-case basis, along with some method to strip identifying information.

benthayer:
This thread was speculative in terms of what we think will be useful. It's better to wait until we know how we can make something useful, so I'm closing this.

Issue #257, opened 2020-08-05
Level: Amending 3 - "The Same Commit"


benthayer:
```
3) The Same Commit - Use branches to show that the commit hash changes

git gud status
--------------------
Before (branch):
Hash: 1234
Message: Commit

After (master):
Hash: 4321
Message: New Commit

======= Simulating: git log --oneline --all --graph
Blah
<<<<<<<
--------------------


A) Starting tree
Initial ---- Commit (master)

B) Branch
Initial ---- Commit (master, branch)

C)
Initial --- New Commit (master)
    \
   Commit (branch)

D)
Initial --- New Commit ---- Merge (master)
    \                       /
   Commit (branch) --------
```

sahansk2:
@benthayer Voop

PR #309, opened 2020-08-30
Added Status Message for Amending

### LINKED ISSUE #264: Amending Status Message
#255, #256, #257 all require a similar status message. The three are listed below:

From #257, there is the most detailed message. In levels #255 and #256, the branches and the hashes aren't there
```
Before (branch):
Hash: 1234
Message: Commit

After (master):
Hash: 4321
Message: New Commit
```

sahansk2:
Fixes #264

Tested locally on skill 2-1
```

(base) sahan@sahan-GA-78LMT-USB3:test$ git gud status
Before:
Hash :  8bce9d1
Message :  Commit 2
File :  Present

After:
Hash :  10c914a
Message :  now i am become death, destroyer of worlds
File :  Missing

### CODE REVIEW

File: gitgud/skills/user_messages.py [hunk 1]
```diff
@@ -177,3 +177,26 @@ def display_entry(index, human_name, code_name, indent):
...
+        snapshot["Message"] = commit.message
+        if show_hashes:
+            snapshot["Hash"] = commit.hexsha[:7]
+        if show_files:
+            files = file_operator.get_commit_content(commit)
+            snapshot["File"] = "Present" if files else "Missing"
+
+    for snapshot in display_data:
+        print(snapshot["_name"] + ":")
+        for feature in ["Hash", "Message", "File"]:
```
benthayer, line 200: Won't this print all three features no matter what? If any of those are not set, this method crashes. Maybe make this dynamic?

gitgud/skills/user_messages.py
```diff
@@ -177,3 +177,26 @@ def display_entry(index, human_name, code_name, indent):
                 code_name=item.name,
                 indent=indent
             )
+
+
+def amending_message(before_ref, after_ref, show_hashes=True, show_files=True, show_refs=True):  # noqa: E501
+    file_operator = operations.get_operator()
+    display_data = [
+        {"_name": "Before", "_commit": file_operator.repo.commit(before_ref)},
+        {"_name": "After", "_commit": file_operator.repo.commit(after_ref)}
```
benthayer, line 186: Here, it might be worth it to do tuples or nested dictionaries. Using underscores is a bit weird here.

gitgud/skills/user_messages.py
```diff
@@ -177,3 +177,26 @@ def display_entry(index, human_name, code_name, indent):
                 code_name=item.name,
                 indent=indent
             )
+
+
+def amending_message(before_ref, after_ref, show_hashes=True, show_files=True, show_refs=True):  # noqa: E501
```
benthayer, line 182: Method doesn't let you show the branches as shown in #257

benthayer (changes requested):
I'm assuming that adding branches would change the method signature. 

To avoid a headache, we can implement that before merging, but if the headache of refactoring later is acceptable, this functionality is still useful and can be merged after the other issues are addressed

sahansk2:
@benthayer The branch displaying support is actually from #310 -- I took the code from there and implemented it here, but now that I think of it, we would have different commits with the same "content". If you're okay with it, I can remove the branch-displaying code from these commits, work on everything else *not* including the branches, and then rebase on master when #310 is merged. What do you think?

PR #310, opened 2020-08-30
Implement Basics Status Message

### LINKED ISSUE #267: Status Method for Basics 2-8
All basic levels in #238 (bar one) use the same or similar status messages to communicate information about commits in the level.

 - #247
 - #248
 - #253
 - #249
 - #261
 - #250
 - #251

The status method should be worked on before any of the levels so all levels can use it.

Can configure
 - Show "Exists" vs Content
 - Show branch
 - Show working directory and staging area or not
 - Potentially more
```
"Initial commit":
 sahan.txt: Exists
 File 2: Doesn't exist
```
```
"Hello world commit" (master):
 sahan.txt: "Hola mundo"
 ben.txt: "Wiggle my toes"
```

sahansk2:
Fixes #267

### DIFF
--- a/gitgud/operations.py
+++ b/gitgud/operations.py
@@ -18,19 +18,16 @@ from gitgud.skills.user_messages import mock_simulate, print_info
 from gitgud.hooks import all_hooks
 
 
-class DirectoryContent:
-    def __init__(self, content):
-        self.content = content
-
+class DirectoryContent(dict):
     def __contains__(self, filepath):
         if isinstance(filepath, Path):
             filepath = str(filepath.as_posix())
-        return filepath in self.content
+        return super().__contains__(filepath)
 
     def __getitem__(self, filepath):
         if isinstance(filepath, Path):
             filepath = str(filepath.as_posix())
-        return self.content[filepath]
+        return super().__getitem__(filepath)
 
 
 def normalize_commit_arg(commit_func):
--- a/gitgud/operations.py
+++ b/gitgud/operations.py
@@ -475,7 +475,7 @@ class Operator():
         for branch_name in tree['branches']:
             target = tree['branches'][branch_name]['target']
             if target not in referred_by:
-                referred_by.update({target: [branch_name]})
+                referred_by[target] = [branch_name]
             else:
                 referred_by[target].append(branch_name)
         return referred_by
--- a/gitgud/operations.py
+++ b/gitgud/operations.py
@@ -469,9 +469,8 @@ class Operator():
                 known_commits[commit_hash] = name
         return known_commits
 
-    def target_branch_str():
-        file_operator = operations.get_operator()
-        tree = file_operator.get_current_tree()
+    def target_branch_str(self):
+        tree = self.get_current_tree()
         referred_by = {}
         for branch_name in tree['branches']:
             target = tree['branches'][branch_name]['target']
@@ -509,6 +508,15 @@ class Operator():
                 mapping[commit_hash] = diffs[diff]
 
         return mapping
+
+    def get_all_commits(self):
+        yielded = set()
+        for head in self.repo.heads:
+            for commit in self.repo.iter_commits(head, reverse=True):
+                if commit not in yielded:
+                    yield commit
+                else:
+                    yielded.add(commit)
 
 
 def get_operator():
--- a/gitgud/operations.py
+++ b/gitgud/operations.py
@@ -13,7 +13,7 @@ from git.exc import InvalidGitRepositoryError
 from gitgud import actor
 from gitgud import skills
 
-from gitgud.skills.user_messages import mock_simulate, print_info
+from gitgud.user_messages import mock_simulate, print_info
 
 from gitgud.hooks import all_hooks
 
--- a/gitgud/operations.py
+++ b/gitgud/operations.py
@@ -469,7 +469,7 @@ class Operator():
                 known_commits[commit_hash] = name
         return known_commits
 
-    def commits_targeted_by(self):
+    def get_branches_by_commit(self):
         tree = self.get_current_tree()
         referred_by = {}
         for branch_name in tree['branches']:
--- a/gitgud/operations.py
+++ b/gitgud/operations.py
@@ -469,7 +469,7 @@ class Operator():
                 known_commits[commit_hash] = name
         return known_commits
 
-    def target_branch_str(self):
+    def commits_targeted_by(self):
         tree = self.get_current_tree()
         referred_by = {}
         for branch_name in tree['branches']:
--- a/gitgud/operations.py
+++ b/gitgud/operations.py
@@ -509,14 +509,15 @@ class Operator():
 
         return mapping
 
-    def get_all_commits(self):
-        yielded = set()
+    def get_all_commits(self, sort_commits=True):
+        all_commits = []
         for head in self.repo.heads:
             for commit in self.repo.iter_commits(head, reverse=True):
-                if commit not in yielded:
-                    yield commit
-                else:
-                    yielded.add(commit)
+                if commit not in all_commits:
+                    all_commits.append(commit)
+        if sort_commits:
+            all_commits.sort(key=lambda commit: commit.committed_date)
+        return all_commits
 
 
 def get_operator():
--- a/gitgud/operations.py
+++ b/gitgud/operations.py
@@ -469,6 +469,18 @@ class Operator():
                 known_commits[commit_hash] = name
         return known_commits
 
+    def target_branch_str():
+        file_operator = operations.get_operator()
+        tree = file_operator.get_current_tree()
+        referred_by = {}
+        for branch_name in tree['branches']:
+            target = tree['branches'][branch_name]['target']
+            if target not in referred_by:
+                referred_by.update({target: [branch_name]})
+            else:
+                referred_by[target].append(branch_name)
+        return referred_by
+
     def get_diffs(self, known_commits):
         diffs = {}
         for commit_hash, commit_name in known_commits.items():
--- a/gitgud/skills/user_messages.py
+++ b/gitgud/skills/user_messages.py
@@ -6,6 +6,12 @@ from .util import Skill
 
 user_has_seen_messages = False
 
+def existence_str(condition):
+    if condition:
+        return "Exists"
+    else:
+        return "Doesn't exist"
+
 
 def start_marker():
     return '<' * 7
@@ -179,29 +185,8 @@ def show_skill_tree(items, show_human_names=True, show_code_names=True, expand_s
             )
 
 
-def target_branch_str():
-    file_operator = operations.get_operator()
-    tree = file_operator.get_current_tree()
-    referred_by = {}
-    for branch_name in tree['branches']:
-        target = tree['branches'][branch_name]['target']
-        if target not in referred_by:
-            referred_by.update({target: [branch_name]})
-        else:
-            referred_by[target].append(branch_name)
-    for target in referred_by:
-        referred_by[target] = ", ".join(referred_by[target])
-    return referred_by
-
-
 def display_tree_data(header, data, show_content, file_count=2):
     file_format_str = "  {path} - {content}"
-
-    def existence_str(condition):
-        if condition:
-            return "Exists"
-        else:
-            return "Doesn't exist"
 
     data_paths = list(data.keys())
     # If there are more files than we requested, we want to show that too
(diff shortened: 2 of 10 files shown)


gitgud/skills/user_messages.py
```diff
@@ -179,7 +179,7 @@ def display_entry(index, human_name, code_name, indent):
             )
 
 
-def basics_status(show_content=True, show_branches=True, working=True, staging=True, file_count=2):
+def basics_status(show_content=True, show_branches=True, working=True, staging=True, file_count=2):  # noqa: E501
```
benthayer, line 182: It might make sense to split this method up.

Instead of passing working/staging, would we be able to pass the data itself? We can then have separate methods for working/staging that simply wrap this function, grabbing the appropriate data

gitgud/skills/user_messages.py
```diff
@@ -219,13 +219,21 @@ def content_display_data(data):
...
 
     # Iterate through heads to get a consistent output.
+    if working:
+        print("Working Directory:")
+        working_data = file_operator.get_working_directory_content()
+        content_display_data(working_data)
+    if staging:
+        print("Staging Directory:")
+        staging_data = file_operator.get_staging_content()
+        content_display_data(staging_data)
```
benthayer, line 236: This part seems isolated, so maybe they can go in their own function?

gitgud/operations.py
```diff
@@ -18,19 +18,16 @@
 from gitgud.hooks import all_hooks
 
 
-class DirectoryContent:
-    def __init__(self, content):
-        self.content = content
-
+class DirectoryContent(dict):
```
benthayer, line 21: I like that we're now subclassing. What additional features do we get from this?

benthayer (commented):
Just commenting since this is still a draft, but hopefully this moves things along. I didn't test or read the code too thoroughly, so do what you feel is best

gitgud/operations.py
sahansk2: The reason why I wanted to do this is because I used the `keys()` method from when `get_*_content` returned a Dictionary for the `content_display_data` method. We can also use anything else that a Dict object could use.

gitgud/skills/user_messages.py
sahansk2: I like the idea of splitting this function up, but I'm wondering how we'd configure `show_branches`, since that parameter only makes sense for commits and not for the working/staging areas.

gitgud/skills/user_messages.py
sahansk2: See my [previous comment](https://github.com/benthayer/git-gud/pull/310#discussion_r485926835).

gitgud/skills/user_messages.py
benthayer: I think there are several ways. We can set `branch=None` and disregard it.  If someone sets it when they're given working directory/staging content, that's their fault :P You can do it whichever way makes sense.

I think I would have a method `show_stuff(name, directory_content, branch_name=None, file_count=2)` and use that to separate the display function from the other parts of this function. `name` would be `"Working Directory"` etc

gitgud/operations.py
benthayer: Are there any potential issues with this? Would we possibly want to override assigning so that this class automatically can convert paths to keys `dir_content[Path('stuff')] = "Hello world"` can be looked up such that `dir_content['stuff'] == "Hello world"`

sahansk2: Hmm, that would be an issue in an abstract sense, but I don't see a benefit of assigning to DirectoryContent. Since the object only serves as a window into the content of the working directory/staging area/commits, DirectoryContent seems only useful in a read-only sense.

That being said, since we're not *assigning* to DirectoryContent values yet (since we're only using it to check pre-existing file content right now), implementing a `__setitem__` that converts Paths to str would be relatively harmless at this point. So, up to you if you think it's worth implementing

benthayer: I think it's unnecessary. If it comes up later, we can add it.

sahansk2:
Strange, I thought I posted a comment...
I split everything into multiple different functions -- `target_branch_str` can probably be moved into Operator, but display_tree_content and display_commit_content together make up the "basics status message". We can directly invoke `display_tree_content` for working/staging, and we can use `display_commit_content` for displaying all commits.

gitgud/skills/user_messages.py
```diff
@@ -177,3 +177,74 @@ def display_entry(index, human_name, code_name, indent):
...
+    commit_format_str = "{message}"
+    if show_branches:
+        commit_format_str += " {branches}"
+    commit_format_str += ":"
+
+    displayed = set()
+    for head in file_operator.repo.heads:
+        for commit in file_operator.repo.iter_commits(head, reverse=True):
+            if commit not in displayed:
+                if commit.hexsha in referred_by and show_branches:
```
benthayer, line 237: I'd reverse these conditions because show_branches is more important.

`show_branches and commit.hexsha in referred_by`

gitgud/skills/user_messages.py
```diff
@@ -177,3 +177,74 @@ def display_entry(index, human_name, code_name, indent):
...
+        if file_exists:
+            path = data_paths[index]
+        else:
+            path = "File " + str(index + 1)
+        if show_content and file_exists:
+            content = data[data_paths[index]]
+        else:
+            content = existence_str(file_exists)
+        print(file_format_str.format(path=path, content=content))
+
```
benthayer, line 222:
```
for commit in op.get_all_commits():
    display_commit_content(commit, *args)
```

gitgud/skills/user_messages.py
```diff
@@ -177,3 +177,74 @@ def display_entry(index, human_name, code_name, indent):
...
+            referred_by[target].append(branch_name)
+    for target in referred_by:
+        referred_by[target] = ", ".join(referred_by[target])
+    return referred_by
+
+
+def display_tree_data(header, data, show_content, file_count=2):
+    file_format_str = "  {path} - {content}"
+
+    def existence_str(condition):
```
benthayer, line 200: This can be in the module scope

gitgud/skills/user_messages.py
```diff
@@ -177,3 +177,74 @@ def display_entry(index, human_name, code_name, indent):
...
+        print(file_format_str.format(path=path, content=content))
+
+
+def display_commit_content(show_branches=True, show_content=True, file_count=2):  # noqa: E501
+    file_operator = operations.get_operator()
+    referred_by = target_branch_str()
+
+    commit_format_str = "{message}"
+    if show_branches:
+        commit_format_str += " {branches}"
```
benthayer, line 230: Parenthesis here around branch

gitgud/skills/user_messages.py
```diff
@@ -177,3 +177,74 @@ def display_entry(index, human_name, code_name, indent):
...
+    commit_format_str += ":"
+
+    displayed = set()
+    for head in file_operator.repo.heads:
+        for commit in file_operator.repo.iter_commits(head, reverse=True):
+            if commit not in displayed:
+                if commit.hexsha in referred_by and show_branches:
+                    branches = "({})".format(referred_by[commit.hexsha])
```
benthayer, line 238: Remove parenthesis here

gitgud/skills/user_messages.py
```diff
@@ -177,3 +177,74 @@ def display_entry(index, human_name, code_name, indent):
...
+                header = commit_format_str.format(
+                    message=commit.message,
+                    branches=branches
+                )
+                display_tree_data(
+                    header,
+                    file_operator.get_commit_content(commit),
+                    file_count=2
+                )
+                displayed.add(commit)
```
benthayer, line 250: Unnecessary if this method is being split up

gitgud/skills/user_messages.py
```diff
@@ -177,3 +177,74 @@ def display_entry(index, human_name, code_name, indent):
...
+        if target not in referred_by:
+            referred_by.update({target: [branch_name]})
+        else:
+            referred_by[target].append(branch_name)
+    for target in referred_by:
+        referred_by[target] = ", ".join(referred_by[target])
+    return referred_by
+
+
+def display_tree_data(header, data, show_content, file_count=2):
```
benthayer, line 197: Rename `data` to `content`

gitgud/skills/user_messages.py
```diff
@@ -177,3 +177,74 @@ def display_entry(index, human_name, code_name, indent):
...
+    file_format_str = "  {path} - {content}"
+
+    def existence_str(condition):
+        if condition:
+            return "Exists"
+        else:
+            return "Doesn't exist"
+
+    data_paths = list(data.keys())
```
benthayer, line 206: file_paths?

gitgud/skills/user_messages.py
```diff
@@ -177,3 +177,74 @@ def display_entry(index, human_name, code_name, indent):
...
+    # If there are more files than we requested, we want to show that too
+
+    print(header + ":")
+    for index in range(max(file_count, len(data_paths))):
```
benthayer, line 210:
```
for path, content in content.items():
    files
for blah2 in range(len(data_paths), file_count):
    not files
```

sahansk2:
@benthayer This is ready for review again. Feel free to get rid of commit 8a3585fa360f176d1c3177944a9227f8900e2f49 if it seems out of place, it's just there to make sure that the functions don't unexpectedly crash during development.

gitgud/operations.py
```diff
@@ -469,6 +469,17 @@ def get_known_commits(self):
                 known_commits[commit_hash] = name
         return known_commits
 
+    def commits_targeted_by(self):
+        tree = self.get_current_tree()
+        referred_by = {}
+        for branch_name in tree['branches']:
```
benthayer, line 475: Why not do `for branch in self.repo.branches`? It would be useful below

gitgud/operations.py
```diff
@@ -469,6 +469,17 @@ def get_known_commits(self):
                 known_commits[commit_hash] = name
         return known_commits
 
+    def commits_targeted_by(self):
```
benthayer, line 472: What do you think of the name `get_branches_by_commit` instead?

gitgud/operations.py
```diff
@@ -498,6 +509,15 @@ def get_copy_mapping(self, non_merges, known_commits):
 
         return mapping
 
+    def get_all_commits(self):
```
benthayer, line 512: Take a look at the code for `get_current_tree`. It seems like we can delete that first while loop and replace the second while loop with `for commit in file_operator.get_all_commits:`

gitgud/operations.py
```diff
@@ -498,6 +509,15 @@ def get_copy_mapping(self, non_merges, known_commits):
...
+        yielded = set()
+        for head in self.repo.heads:
+            for commit in self.repo.iter_commits(head, reverse=True):
```
benthayer, line 515: This gets us all the branches, but it will miss any commits if there's a detached head or tags where the commits aren't present. You should also add a `sorted` flag in case you want the commits sorted (which I think you do)

gitgud/skills/user_messages.py
```diff
@@ -179,46 +186,25 @@ def display_entry(index, human_name, code_name, indent):
...
-        if file_exists:
-            path = data_paths[index]
-        else:
-            path = "File " + str(index + 1)
-        if show_content and file_exists:
-            content = data[data_paths[index]]
-        else:
-            content = existence_str(file_exists)
-        print(file_format_str.format(path=path, content=content))
+    for filepath in available_files:
```
benthayer, line 196: You can just use `content.keys()` here

gitgud/skills/user_messages.py
```diff
@@ -179,46 +186,25 @@ def display_entry(index, human_name, code_name, indent):
...
     file_format_str = "  {path} - {content}"
 
-    def existence_str(condition):
-        if condition:
-            return "Exists"
-        else:
-            return "Doesn't exist"
-
-    data_paths = list(data.keys())
+    available_files = list(content.keys())
```
benthayer, line 192: You don't need to do this

gitgud/skills/user_messages.py
```diff
@@ -179,46 +186,25 @@ def display_entry(index, human_name, code_name, indent):
...
-            path = data_paths[index]
-        else:
-            path = "File " + str(index + 1)
-        if show_content and file_exists:
-            content = data[data_paths[index]]
-        else:
-            content = existence_str(file_exists)
-        print(file_format_str.format(path=path, content=content))
+    for filepath in available_files:
+        file_content = filepath if show_content else existence_str(True)
```
benthayer, line 197: Mistake here, it should be `file_content = content[filepath] ...`

Also, the name `file_content` is a bit misrepresentative since it doesn't show the content. Maybe `file_details`/`file_info` or something more encapsulating? If you do that, can you similarly change the name in the format string also

benthayer (changes requested):

benthayer:
Should be all good. I still want to refactor the user messages:
```
gitgud.skills.user_messages.py -> gitgud.__init__.py
gitgud.skills.state_messages.py -> gitgud.stateful.py
```

Issue #238, opened 2020-07-16
New Basic Levels


benthayer:
The idea is to replace basics 1 through 3

```
1) First Commit - Show that you can add/commit - Commits contain files (adapt from basics-committing)
2) Two Commits - Commits are sequential - Branch updates as you commit

--------------------
working directory:
 File 1: Exists
 File 2: Exists
staging area:
 File 1: Exists
 File 2: Exists
commit 1: Doesn't exist
commit 2: Doesn't exist
--------------------

--------------------
"Initial commit":
 File 1: Exists
 File 2: Doesn't exist
"Hello world commit" (master):
 File 1: Exists
 File 2: Exists
--------------------
3) Five Commits - add/rm/modify/mv/multiple changes at once

gg goal
add file 1
add file 2
modify file 2
remove file 1
move file 2

gg status
--------------------
commit 1:
 sahan.txt: "Hello world"
 File 2: Doesn't exist
commit 2:
 sahan.txt: "Hello world"
 File 2: "Barf on my feet"
commit 3:
 File 1: "Hello world"
 File 2: "Put nail polish on my toes"
commit 4:
 File 1: Doesn't exist
 File 2: "Put nail polish on my toes"
commit 4: (master)
 File 1: Doesn't exist
 File 3: "Put nail polish on my toes"
-------------------

4) Branch/Merge (to be adapted from basics-merging)

1
2 (branch)
3 : 1 (master)
M : 2 3 (master but later)

gg solution
git checkout master
git merge branch

--------------------
wd
staging
commit 1
 A
commit 2 (branch)
 A
 B
commit 3 (master)
 A
 C
merge (master but later)
 A
 B
 C
--------------------

5) Merge Conflicts

--------------------
wd
staging
 - omg a merge (at some point)
commit 1
 A
commit 2 (branch)
 A
 B
commit 3 (master)
 A
 B'
merge (master but later)
 A
 B or B'???: "I'm the first line...."
--------------------


6) DIY Merge Conflicts (partial adaptation from basics-branching)

Literally the same as the last level, but you do it yourself
```

sahansk2:
I'll tackle this issue.

sahansk2:
I think that tackling this would depend on #224 being resolved first.

benthayer:
Yes, one assumption is that we can name commits whatever we'd like. It also depends on us being able to set the content of the files within the commit. I can add issues for those later. 

Do you feel comfortable resolving this? It definitely gets into the nitty gritty

PR #314, opened 2020-09-08
Add testing for methods defined in GitGud object

### LINKED ISSUE #197: Check error codes in test suites
When commands fail, they should fail with status code 1, which could help us make sure users are never seeing an error that's coming from python.

sahansk2:
Fixes #197 

@benthayer Could you verify for which methods you would like tests?


gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
+import pytest
+
+from gitgud.__main__ import GitGud
+
+def test_assert_initialized():
+    pass
+def test_handle_commit():
+    pass
+def test_handle_contrib():
```
benthayer, line 9: No test

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
...
+    pass
+def test_handle_debug():
```
benthayer, line 11: No test

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
+import pytest
+
+from gitgud.__main__ import GitGud
+
+def test_assert_initialized():
```
benthayer, line 5: Test in git gud repo, git repo, empty directory

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
...
+    pass
+def test_handle_skills():
+    pass
+def test_handle_solution():
+    pass
+def test_handle_status():
+    pass
+def test_handle_test():
+    pass
+def test_is_initialized():
```
benthayer, line 43: Test with assert_initilized

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
+import pytest
+
+from gitgud.__main__ import GitGud
+
+def test_assert_initialized():
+    pass
+def test_handle_commit():
```
benthayer, line 7: Since we might be removing this, no test yet unless you want to

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
...
+    pass
+def test_handle_contrib():
+    pass
+def test_handle_debug():
+    pass
+def test_handle_explain():
```
benthayer, line 13: Ideally, we test explain/goal/status on all levels and at all stages during the solving, but to keep things lightweight, let's maybe just do it for one level as a sanity check?

We could run the commands on all levels, but that is more time-consuming. What are your thoughts?

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
...
+    pass
+def test_handle_goal():
+    pass
+def test_handle_help():
+    pass
+def test_handle_init():
+    pass
+def test_handle_issues():
```
benthayer, line 21: No test

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
...
+    pass
+def test_handle_contrib():
+    pass
+def test_handle_debug():
+    pass
+def test_handle_explain():
+    pass
+def test_handle_goal():
+    pass
+def test_handle_help():
```
benthayer, line 17: No test

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
...
+    pass
+def test_handle_init():
```
benthayer, line 19: Test in empty dir/ non-empty dir/ init-ed dir

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
...
+    pass
+def test_handle_issues():
+    pass
+def test_handle_level():
+    pass
+def test_handle_levels():
```
benthayer, line 25: level/levels/skill - test in initilized dir/ uninit-ed dir

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
...
+    pass
+def test_handle_load():
```
benthayer, line 27: test load next/prev and test loading some generic level (or levels if you want).

Also test outside of an init-ed dir to make sure it doesn't throw an error

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
...
+    pass
+def test_handle_reset():
+    pass
+def test_handle_show():
+    pass
+def test_handle_show_tree():
```
benthayer, line 33: Test output in git dir/non-git dir

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
...
+    pass
+def test_handle_level():
+    pass
+def test_handle_levels():
+    pass
+def test_handle_load():
+    pass
+def test_handle_reset():
+    pass
+def test_handle_show():
```
benthayer, line 31: Test with show-tree

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
...
+    pass
+def test_handle_show_tree():
+    pass
+def test_handle_skills():
+    pass
+def test_handle_solution():
```
benthayer, line 37: Same tests as other user facing commands, but obviously this one, you'll need to test the --confirm flag

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
...
+    pass
+def test_handle_status():
+    pass
+def test_handle_test():
```
benthayer, line 41: Load up a level and make sure that it it doesn't fail either when the level has been completed or when it hasn't

Also make sure it doesn't fail if we're not in a git dir

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
...
+    pass
+def test_is_initialized():
+    pass
+def test_load_level():
+    pass
+def test_parse():
```
benthayer, line 47: Test `git gud` without any subcommands in git dir and non-git dir

gitgud/test_user.py
```diff
@@ -0,0 +1,48 @@
...
+    pass
+def test_handle_solution():
+    pass
+def test_handle_status():
+    pass
+def test_handle_test():
+    pass
+def test_is_initialized():
+    pass
+def test_load_level():
```
benthayer, line 45: `commits.csv` should be modified after this func is run, and the level file should correspond to the one loaded.

There should be a test for this in `test_operator.py` if there isn't one already

benthayer (commented):
I was a bit quick with the comments here to try and be expedient. I think everything is covered. Let me know if you need clarification.

Also, at this point, I think creating levels is the priority, not necessarily the tests, so don't feel like you're obligated to write all these tests

benthayer:
It might (probably will) make sense to split this up into multiple PRs

sahansk2:
If you'd like multiple PRs, would you also like multiple issues for this? E.g. "Test user methods using webbrowser", "Test user methods related to loading", "Test assert initialized methods", etc?

benthayer:
Yes, I think that makes sense. Obviously, there would be many issues, and writing the issues might take as long as it takes to simply write the test, so do what you feel is right. There should either be a pull request or an issue for each test

sahansk2:
What we could do is make a new "meta-issue", similar to what #238 is. Each checkbox could be a method in the GitGud object.

Going off of this, we could maintain another branch `user_tests` on this repository, and people open up testing PRs to merge to `user_tests`. Once all the tests are in and everything looks set, we merge `user_tests` into `master`. This is similar to what the `intro` branch (or the `add` branch, I forget) was like a while back.

I'm not sure what's best, so it's your call. As an aside, should this PR be closed?

benthayer:
I think we can open a meta-issue (along with other issues)

I don't think we need to make a separate branch for the tests, I think we can merge them right into master when they are done. 

This PR can stay open until there exists other issues/PRs that make use of the useful comments that are here

benthayer:
Stale/closing

PR #290, opened 2020-08-15
Level: Rebasing Instead

benthayer:
Fixes #261

PR #321, opened 2020-09-13
Fix errors setting up levels during rebases and bisects

benthayer:
During a rebase conflict or a bisect, the head branch would be considered checked out. Since a checked out branch couldn't be deleted, we had trouble setting up new levels during resets/loads/re-inits. This should fix that.

#290 Needs this


benthayer:
@sahansk2 Would you mind taking a look at this?

gitgud/test_operations.py
```diff
@@ -0,0 +1,56 @@
...
+        # This will happen every time
+        pass
+
+
+def setup_bisect():
+    file_operator = get_operator()
+    file_operator.repo.git.bisect('start')
+
+
+def test_reset_during_rebase_conflict(gg):
```
sahansk2, line 33: Minor nitpick, I'd change the testing to use fixtures instead, but that's not the point of this PR. So, feel free to keep it as it is, if you want.

sahansk2 (approved):
Yep, looks good! Just added a [minor] comment about testing.

gitgud/test_operations.py
benthayer: We got a pytest

PR #328, opened 2020-11-26
Refactored subparsers

benthayer:
The goal here is to make the parsers a whole lot easier to comprehend


benthayer:
@sahansk2 I know you haven't touched this in a while, but I was feeling like the commands were getting unwieldy. Do you think you can take a look at this?

benthayer:
I was thinking about doing a loop in `__init__` to make the arguments automatically include the self variable, so you'd only have to call `args.func(args)` in `GitGud.parse`, but that seems like overkill and I'm not sure how to do it well

